PR #4, opened 2018-06-27
fix rosdep key for pytest

dirk-thomas:


dirk-thomas:
Devel job passing now: [Build Status](http://build.ros2.org/job/Bdev__rosidl_python__ubuntu_bionic_amd64/5/)

PR #17, opened 2018-10-31
Allow generated IDL files

sloretz:
The target dependency won't exist at cmake configure time before the first build if it is generated. Since the custom command is in a subdirectory it cannot depend on the generated files directly. Instead it depends on a target from the parent directory which depends on the IDL files.

connects to ros2/rcl_interfaces#47


jacobperron:
Needs a rebase on master.

sloretz:
@hidmic Added comment in 1bf014631789dfbf42537c91659419eb9260e8e2

Issue #16, opened 2018-10-29
Add support to get message field types


fkromer:
In ROS1 it's possible to get the message fields (here: `'__slots__': ['x', 'y', 'z']`) and their corresponding types (here: `'_slot_types': ['float64', 'float64', 'float64']` via `_get_types()`) from messages (here: `geometry_msgs.msg.Point`).

```
$ ipython
Python 2.7.6 (default, Nov 23 2017, 15:49:48) 
Type "copyright", "credits" or "license" for more information.

IPython 5.5.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: from geometry_msgs.msg import Point

In [4]: Point.__dict__
Out[4]: 
dict_proxy({'__doc__': None,
            '__init__': <function geometry_msgs.msg._Point.__init__>,
            '__module__': 'geometry_msgs.msg._Point',
            '__slots__': ['x', 'y', 'z'],
            '_full_text': '# This contains the position of a point in free space\nfloat64 x\nfloat64 y\nfloat64 z\n',
            '_get_types': <function geometry_msgs.msg._Point._get_types>,
            '_has_header': False,
            '_md5sum': '4a842b65f413084dc2b10fb484ea7f17',
            '_slot_types': ['float64', 'float64', 'float64'],
            '_type': 'geometry_msgs/Point',
            'deserialize': <function geometry_msgs.msg._Point.deserialize>,
            'deserialize_numpy': <function geometry_msgs.msg._Point.deserialize_numpy>,
            'serialize': <function geometry_msgs.msg._Point.serialize>,
            'serialize_numpy': <function geometry_msgs.msg._Point.serialize_numpy>,
            'x': <member 'x' of 'Point' objects>,
            'y': <member 'y' of 'Point' objects>,
            'z': <member 'z' of 'Point' objects>})
$ python
>>> from geometry_msgs.msg import Point
>>> p = Point()
>>> p.__slots__
['x', 'y', 'z']
>>> p._get_types()
['float64', 'float64', 'float64']
```

ROS2 doesn't support this right now.

```
$ ipython3
Python 3.6.5 (default, Apr  1 2018, 05:46:30) 
Type "copyright", "credits" or "license" for more information.

IPython 5.5.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
[...]

PR #19, opened 2018-11-03
Adding a get_slot_fields_and_types method to python msg classes

mlautman:
This adds a `_slot_types` field to python message classes as was requested here: https://github.com/ros2/rosidl_python/issues/16

This attempts to remain as true to the ROS1 functionality as possible

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -1,4 +1,4 @@
-# generated from rosidl_generator_py/resource/_msg.py.em  # noqa: W504
+# generated from rosidl_generator_py/resource/_msg.py.em
 # generated code does not contain a copyright notice
 
 @#######################################################################
@@ -236,26 +236,26 @@ if field.name in dict(inspect.getmembers(builtins)).keys():
 @[  end if]@
             assert \
 @[  if field.type.is_array]@
-                ((isinstance(value, Sequence) or  # noqa: W504
-                  isinstance(value, Set) or  # noqa: W504
-                  isinstance(value, UserList)) and  # noqa: W504
-                 not isinstance(value, str) and  # noqa: W504
-                 not isinstance(value, UserString) and  # noqa: W504
+                ((isinstance(value, Sequence) or
+                  isinstance(value, Set) or
+                  isinstance(value, UserList)) and
+                 not isinstance(value, str) and
+                 not isinstance(value, UserString) and
 @{assert_msg_suffixes = ['a set or sequence']}@
 @[    if field.type.type == 'string' and field.type.string_upper_bound]@
-                 all(len(val) <= @field.type.string_upper_bound for val in value) and  # noqa: W504
+                 all(len(val) <= @field.type.string_upper_bound for val in value) and
 @{assert_msg_suffixes.append('and each string value not longer than %d' % field.type.string_upper_bound)}@
 @[    end if]@
 @[    if field.type.array_size]@
 @[      if field.type.is_upper_bound]@
-                 len(value) <= @(field.type.array_size) and  # noqa: W504
+                 len(value) <= @(field.type.array_size) and
 @{assert_msg_suffixes.insert(1, 'with length <= %d' % field.type.array_size)}@
 @[      else]@
-                 len(value) == @(field.type.array_size) and  # noqa: W504
+                 len(value) == @(field.type.array_size) and
 @{assert_msg_suffixes.insert(1, 'with length %d' % field.type.array_size)}@
 @[      end if]@
 @[    end if]@
-                 all(isinstance(v, @(get_python_type(field.type))) for v in value) and  # noqa: W504
+                 all(isinstance(v, @(get_python_type(field.type))) for v in value) and
 @{assert_msg_suffixes.append("and each value of type '%s'" % get_python_type(field.type))}@
 @[    if field.type.type.startswith('int')]@
 @{
@@ -279,7 +279,7 @@ bound = 2**nbits
 @[    end if]@
                 "The '@(field.name)' field must be @(' '.join(assert_msg_suffixes))"
 @[  elif field.type.string_upper_bound]@
-                ((isinstance(value, str) or isinstance(value, UserString)) and  # noqa W504
+                ((isinstance(value, str) or isinstance(value, UserString)) and
                  len(value) <= @(field.type.string_upper_bound)), \
                 "The '@(field.name)' field must be string value " \
                 'not longer than @(field.type.string_upper_bound)'
@@ -287,11 +287,11 @@ bound = 2**nbits
                 isinstance(value, @(field.type.type)), \
                 "The '@(field.name)' field must be a sub message of type '@(field.type.type)'"
 @[  elif field.type.type == 'byte']@
-                ((isinstance(value, bytes) or isinstance(value, ByteString)) and  # noqa: W504
+                ((isinstance(value, bytes) or isinstance(value, ByteString)) and
                  len(value) == 1), \
                 "The '@(field.name)' field must of type 'bytes' or 'ByteString' with a length 1"
 @[  elif field.type.type == 'char']@
-                ((isinstance(value, str) or isinstance(value, UserString)) and  # noqa: W504
+                ((isinstance(value, str) or isinstance(value, UserString)) and
                  len(value) == 1 and ord(value) >= -128 and ord(value) < 128), \
                 "The '@(field.name)' field must of type 'str' or 'UserString' " \
                 'with a length 1 and the character ord() in [-128, 127]'
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -1,4 +1,4 @@
-# generated from rosidl_generator_py/resource/_msg.py.em
+# generated from rosidl_generator_py/resource/_msg.py.em  # noqa: W504
 # generated code does not contain a copyright notice
 
 @#######################################################################
@@ -14,6 +14,7 @@
 @#  - value_to_py (function)
 @#######################################################################
 @
+from copy import copy
 import logging
 import traceback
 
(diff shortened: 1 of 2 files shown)


mlautman:
@dirk-thomas

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -114,9 +114,27 @@ class @(spec.base_type.type)(metaclass=Metaclass):
         '_@(field.name)',
 @[end for]@
     ]
+
+    _slot_types = [
+@[for field in spec.fields]@
+@[  if field.type.is_primitive_type() ]@
+@[    if field.type.is_array ]@
+        '@(get_python_type(field.type))[@(field.type.array_size)]',
```
dirk-thomas: This doesn't distinguish between a static array (with a fixed size) and a dynamic array (either with an upper bound or without a bound).

mlautman: True. Do you have a preferred method to denote the different types of arrays? I didn't have a strong intuition here

dirk-thomas: I would suggest either the same notation as in the `.msg` or an object representation. Thinking about it the second one is probably be better in terms of stability with the upcoming change to `.idl` files since the notation will change (see https://github.com/ros2/design/blob/gh-pages/articles/142_idl.md).

mlautman: I know I can use `field.type.array_size.is_upper_bound` for the upper bound arrays but I'm not sure how to check for unbounded arrays. Any hints?

mlautman:
@dirk-thomas ready for another review

rosidl_generator_py/resource/_msg.py.em
dirk-thomas: Looking at the patch I don't see where the type is being distinguished?

mlautman: My apologies, I got my branches mixed up. All of my changes are in one commit (b9a6a81abc8f75fec9899eb0e9c71b889b448620) but I'm unsure of whether I should cherry pick that commit on top of master or 0.5.2. Since this is targeting master, I assume that I should cherry pick on top of that, but doing so causes the build to fail for unrelated reasons..  

`/home/mike/ws_rqt2/build/rosidl_generator_py/rosidl_generator_py/rosidl_generator_py/msg/_string_arrays_s.c:12:10: fatal error: rosidl_generator_c/primitives_sequence.h: No such file or directory
 #include <rosidl_generator_c/primitives_sequence.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
/home/mike/ws_rqt2/build/rosidl_generator_py/rosidl_generator_py/rosidl_generator_py/msg/_nested_s.c:12:10: fatal error: rosidl_generator_c/primitives_sequence.h: No such file or directory
 #include <rosidl_generator_c/primitives_sequence.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
/home/mike/ws_rqt2/build/rosidl_generator_py/rosidl_generator_py/rosidl_generator_py/msg/_various_s.c:12:10: fatal error: rosidl_generator_c/primitives_sequence.h: No such file or directory
 #include <rosidl_generator_c/primitives_sequence.h>
`

Is there a .repos available with the most recent ros2 branches that I should use for making PR's to master?

mlautman: @dirk-thomas

dirk-thomas: It depends on what you are trying to build. I would suggest using the `.repos` files from the `master` branch in the `ros2/ros2` repo and building everything from source for your use case. Then a patch based on the master of this repo will work.

(The opposite would be to use the `bouncy` branch of the `.repos` file and then also target the `bouncy` branch in this repo. But usually all contributions are targeting the default branch and only in case of severe bug fixes are being considered for backporting afterwards.)

dirk-thomas: The current state seems to pass on CI: [Build Status](https://ci.ros2.org/job/ci_linux/5548/)

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -114,9 +114,31 @@ class @(spec.base_type.type)(metaclass=Metaclass):
         '_@(field.name)',
 @[end for]@
     ]
+
+    _slot_types = [
+@[for field in spec.fields]@
+@[  if field.type.is_primitive_type() ]@
+        '@(field.type)',
```
dirk-thomas: A string could still have an upper boundary.

dirk-thomas: The string representation of primitives contains these parts already.

mlautman:
@dirk-thomas Changed `_slot_types_dict` to `get_slot_types_dict()` and added noqa W504 where appropriate to pass flake8

mlautman:
https://ci.ros2.org/job/ci_linux/5564/

dirk-thomas:
Ignoring W504 all over the place isn't necessary. This has been addressed in the linter configuration already a while ago: ament/ament_lint#110.

mlautman:
@dirk-thomas It still fails locally with `colcon build test --package-select rosidl_python` without that change

dirk-thomas:
> It still fails locally

Are you using the latest default branches of all the other ROS 2 repos?

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -175,6 +197,10 @@ class @(spec.base_type.type)(metaclass=Metaclass):
             return False
 @[end for]@
         return True
+
+    @@classmethod
+    def get_slot_types_dict(cls):
```
mlautman:
```suggestion
    def get_slot_fields_and_types(cls):
```

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -257,3 +257,51 @@ def test_out_of_range():
         setattr(b, 'unbounded_uint64_values', [2**64])
     with pytest.raises(AssertionError):
         setattr(b, 'unbounded_uint64_values', [-1])
+
+
+def test_slot_attributes():
+    a = Nested()
+    assert hasattr(a, 'get_slot_types_dict')
+    assert hasattr(a, '__slots__')
+    nested_slot_types_dict = getattr(a, 'get_slot_types_dict')()
```
mlautman:
```suggestion
    nested_slot_types_dict = getattr(a, 'get_slot_fields_and_types')()
```

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -257,3 +257,51 @@ def test_out_of_range():
         setattr(b, 'unbounded_uint64_values', [2**64])
     with pytest.raises(AssertionError):
         setattr(b, 'unbounded_uint64_values', [-1])
+
+
+def test_slot_attributes():
+    a = Nested()
+    assert hasattr(a, 'get_slot_types_dict')
```
mlautman:
```suggestion
    assert hasattr(a, 'get_slot_fields_and_types')
```

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -257,3 +257,51 @@ def test_out_of_range():
...
+    assert len(nested_slot_types_dict) == len(expected_nested_slot_types_dict)
+
+    for expected_field, expected_slot_type in expected_nested_slot_types_dict.items():
+        assert expected_field in nested_slot_types_dict.keys()
+        assert expected_slot_type == nested_slot_types_dict[expected_field]
+
+
+def test_primative_slot_attributes():
+    b = StringArrays()
+    assert hasattr(b, 'get_slot_types_dict')
```
mlautman:
```suggestion
    assert hasattr(b, 'get_slot_fields_and_types')
```

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -257,3 +257,51 @@ def test_out_of_range():
...
+    assert hasattr(b, '__slots__')
+    string_slot_types_dict = getattr(b, 'get_slot_types_dict')()
```
mlautman:
```suggestion
    string_slot_types_dict = getattr(b, 'get_slot_fields_and_types')()
```

rosidl_generator_py/resource/_msg.py.em
brawner:
```suggestion
    def get_fields_and_field_types(cls):
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -175,6 +197,10 @@ class @(spec.base_type.type)(metaclass=Metaclass):
             return False
 @[end for]@
         return True
+
+    @@classmethod
+    def get_fields_and_field_types(cls):
+        return cls._slot_types_dict
```
brawner: Since this is a getter, should it return a shallow copy so that you can modify the underlying dictionary?

mlautman: Good catch. Fixed.

mlautman:
@dirk-thomas I am using the ros2.repos that you passed along to us with the 0.5.2 release.

I went back and removed W504. I also changed the code to return a copy of the dict rather than the dict itself per @brawner's comment

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -252,7 +279,7 @@ bound = 2**nbits
 @[    end if]@
                 "The '@(field.name)' field must be @(' '.join(assert_msg_suffixes))"
 @[  elif field.type.string_upper_bound]@
-                ((isinstance(value, str) or isinstance(value, UserString)) and
+                ((isinstance(value, str) or isinstance(value, UserString)) and  # noqa W504
```
mlautman:
```suggestion
                ((isinstance(value, str) or isinstance(value, UserString)) and
```

dirk-thomas:
This can be merged as is for now. With the coming change I will likely change this in the following way:

* Store the types in a class constant, e.g. `SLOT_TYPES`.
* Change the data structure from a mutable `dict` to an immutable `tuple` of [abstract types](https://github.com/ros2/rosidl/blob/49b1eaa23eb4d11c73dca7ea610a1cce6f129607/rosidl_parser/rosidl_parser/definition.py#L48)
* Either remove the class method (or rename it to `get_slot_types` which zips `__slots__` and `SLOT_TYPES` and returns an ordered dictionary).

PR #21, opened 2018-11-15
Python typesupport for Actions

### LINKED ISSUE #14: Generate python typesupport for Actions
## Feature request

There should be a python module importable by `import packagename.action.actioname` which contains everything needed to use actions in python.

#### Feature description

The module needs to have references to all the types a user needs to interface with actions. This should follow the same model as services since a service type support includes two message type supports.

#### Implementation considerations

The service python object imports the generated python objects for the request and response message. The action module should import the generated python objects for the services and message it uses as well as the common cancel service and status message.

https://github.com/ros2/rosidl_python/blob/deab237bf2a51735e0a449cc19bb23c28b47dbd5/rosidl_generator_py/resource/_srv.py.em#L35-L41

The python object for services is just a class with two class members: `Request` and `Response`. Actions could be a class with 3 members: `SendGoal`, `GetResult`, and `Feedback`

https://github.com/ros2/rosidl_python/blob/deab237bf2a51735e0a449cc19bb23c28b47dbd5/rosidl_generator_py/resource/_srv.py.em#L44-L46

connects to ros2/ros2#583

apojomovsky:
Closes #14 
This PR:
- Adds the required `rosidl_python/rosidl_generator_py/resource/_action.py.em` template file with the proper imports for the generated action messages/services.
- Adds action typesupport bits to the `_msg_pkg_typesupport_entry_point.c.em` script.
- Adds logic to the `rosidl_generator_py` so that it can handle the new action's template.

### DIFF
--- a/rosidl_generator_py/cmake/custom_command.cmake
+++ b/rosidl_generator_py/cmake/custom_command.cmake
@@ -23,7 +23,7 @@ add_custom_command(
   COMMAND ${PYTHON_EXECUTABLE} ${rosidl_generator_py_BIN}
   --generator-arguments-file "${generator_arguments_file}"
   --typesupport-impls "${_typesupport_impls}"
-  DEPENDS ${target_dependencies} ${rosidl_generate_interfaces_TARGET}
+  DEPENDS ${target_dependencies} ${rosidl_generate_interfaces_TARGET} ${extra_generator_dependencies}
   COMMENT "Generating Python code for ROS interfaces"
   VERBATIM
 )
--- a/rosidl_generator_py/resource/_action.py.em
+++ b/rosidl_generator_py/resource/_action.py.em
@@ -1,0 +1,61 @@
+# generated from rosidl_generator_py/resource/_action.py.em
+# generated code does not contain a copyright notice
+
+@#######################################################################
+@# EmPy template for generating _<action>.py files
+@#
+@# Context:
+@#  - module_name
+@#  - package_name
+@#  - spec (rosidl_parser.ActionSpecification)
+@#    Parsed specification of the .action file
+@#  - convert_camel_case_to_lower_case_underscore (function)
+@#######################################################################
+@
+import logging
+import traceback
+
+
+class Metaclass(type):
+    """Metaclass of action '@(spec.action_name)'."""
+
+    _TYPE_SUPPORT = None
+
+    @@classmethod
+    def __import_type_support__(cls):
+        try:
+            from rosidl_generator_py import import_type_support
+            module = import_type_support('@(package_name)')
+        except ImportError:
+            logger = logging.getLogger('rosidl_generator_py.@(spec.action_name)')
+            logger.debug(
+                'Failed to import needed modules for type support:\n' + traceback.format_exc())
+        else:
+            cls._TYPE_SUPPORT = module.type_support_action__@(subfolder)_@(module_name)
+@{
+preffix = '_' + convert_camel_case_to_lower_case_underscore(spec.action_name) + '__'
+suffixes = ['feedback', 'goal', 'result']
+for field_name in [preffix + suffix for suffix in suffixes]:
+    print('%sfrom %s.%s import %s' % (' ' * 4 * 3, package_name, subfolder, field_name))
+    print('%sif %s.Metaclass._TYPE_SUPPORT is None:' % (' ' * 4 * 3, field_name))
+    print('%s%s.Metaclass.__import_type_support__()' % (' ' * 4 * 4, field_name))
+print('%sfrom %s.%s import %s' % (' ' * 4 * 3, 'action_msgs', 'msg', '_goal_status'))
+print('%sif %s.Metaclass._TYPE_SUPPORT is None:' % (' ' * 4 * 3, '_goal_status'))
+print('%s%s.Metaclass.__import_type_support__()' % (' ' * 4 * 4, '_goal_status'))
+print('%sfrom %s.%s import %s' % (' ' * 4 * 3, 'action_msgs', 'srv', '_cancel_goal'))
+print('%sif %s.Metaclass._TYPE_SUPPORT is None:' % (' ' * 4 * 3, '_cancel_goal'))
+print('%s%s.Metaclass.__import_type_support__()' % (' ' * 4 * 4, '_cancel_goal'))
+}@
+
+class @(spec.action_name)(metaclass=Metaclass):
+    from action_msgs.srv._cancel_goal import CancelGoal as CancelGoalService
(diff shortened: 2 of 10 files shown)


brawner:
The fork this PR is coming from is not accessible. Was it dropped, or just not made public?

apojomovsky:
Hi @brawner , I think the branch name has lead to confusion here. Actually, this PR is coming from a local branch to the rcl repo and can be found here: https://github.com/ros2/rosidl_python/tree/apojomovsky/typesupport_for_actions

brawner:
Haha, that indeed was my confusion. Thanks!

brawner:
I've tested this out on my system. Is there a reason why no high-level action class is created? Something like:

```
from test_msgs.action import Fibonacci
```

as opposed to
```
from test_msgs.action import Fibonacci_Goal, Fibonacci_Result, Fibonacci_Feedback
```

jacobperron:
> I've tested this out on my system. Is there a reason why no high-level action class is created? Something like:
>
>      from test_msgs.action import Fibonacci

It would appear that the `_action.py.em` template is not being invoked. It possibly because of current rosidl pipeline where (I think) `.action` files are being broken into services/messages prior to being passed to the generators. @apojomovsky Can you confirm?

apojomovsky:
Hi @jacobperron , I've just verified and it's exactly as you say. With the current pipeline, action files are broken into services and messages prior to being passed to the generator.

rosidl_generator_py/cmake/custom_command.cmake
```diff
@@ -23,7 +23,8 @@ add_custom_command(
   COMMAND ${PYTHON_EXECUTABLE} ${rosidl_generator_py_BIN}
   --generator-arguments-file "${generator_arguments_file}"
   --typesupport-impls "${_typesupport_impls}"
-  DEPENDS ${target_dependencies} ${rosidl_generate_interfaces_TARGET}
+  ${extra_generator_arguments}
```
dirk-thomas: Where is this variable being defined?

hidmic: A left over from a previous attempt. Thanks for pointing it out. Removed.

hidmic:
Alright, it seems to be working now. Or at least I'm able to import the `Fibonacci` class from `test_msgs.action` and get its C type support with `Fibonacci.__import_type_support__()`.

Having said that, I'm not very proud of this patch. It works, but it's _fragile_, if you know what I mean. The split pipeline does no good to the Python generator implementation, and this is the solution I came up with that minimized the amount of changes necessary. Hopefully, we'll get to replace it with a more robust implementation once Dirk's work lands.

@jacobperron this is ready for you to implement actions in `rclpy` (and catch bugs, if any, that's the ultimate test).

hidmic:
Running CI on this one in the meantime:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6011/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2517/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4961/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5885/)

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -104,23 +118,46 @@ def generate_py(generator_arguments_file, typesupport_impls):
...
+            with open(path_to_module, 'r') as f:
+                content = f.read()
+        with open(path_to_module, 'w') as f:
+            block_name = args['package_name']
+            if action_specs:
+                block_name += '_action'
+            content = re.sub(
+                r"# BEGIN %s$.*^# END %s" % (block_name, block_name),
+                '', content, 0, re.M | re.S
+            )
```
sloretz: Is this part where it modifies the previously generated `__init__.py` the fragile part you were referring to?

rosidl_generator_py/resource/_action.py.em
```diff
@@ -0,0 +1,62 @@
...
+    print('%sif %s.Metaclass._TYPE_SUPPORT is None:' % (' ' * 4 * 3, field_name))
+    print('%s%s.Metaclass.__import_type_support__()' % (' ' * 4 * 4, field_name))
+print('%sfrom %s.%s import %s' % (' ' * 4 * 3, 'action_msgs', 'msg', '_goal_status'))
+print('%sif %s.Metaclass._TYPE_SUPPORT is None:' % (' ' * 4 * 3, '_goal_status'))
+print('%s%s.Metaclass.__import_type_support__()' % (' ' * 4 * 4, '_goal_status'))
+print('%sfrom %s.%s import %s' % (' ' * 4 * 3, 'action_msgs', 'srv', '_cancel_goal'))
+print('%sif %s.Metaclass._TYPE_SUPPORT is None:' % (' ' * 4 * 3, '_cancel_goal'))
+print('%s%s.Metaclass.__import_type_support__()' % (' ' * 4 * 4, '_cancel_goal'))
+}@
+
```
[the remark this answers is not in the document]
hidmic: True. Fixed.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
hidmic: That and adding `_action` to the package name whenever necessary (but not everywhere) to avoid name collisions.

hidmic:
If [this thread](https://cmake.org/pipermail/cmake/2016-April/063150.html) is right, https://github.com/ros2/rosidl_python/pull/21/commits/3e212cb8711a648f55cb6b532f91bdc1142444ff should fix our little issue.

hidmic:
Re-running Windows CI, last one failed for unrelated reasons.

rosidl_generator_py/resource/_action.py.em
```diff
@@ -0,0 +1,62 @@
...
+            from rosidl_generator_py import import_type_support
+            module = import_type_support('@(package_name)', '@(package_name)_action')
+        except ImportError:
+            logger = logging.getLogger('rosidl_generator_py.@(spec.action_name)')
+            logger.debug(
+                'Failed to import needed modules for type support:\n' + traceback.format_exc())
+        else:
+            cls._TYPE_SUPPORT = module.type_support_action__@(subfolder)_@(module_name)
+@{
+preffix = '_' + convert_camel_case_to_lower_case_underscore(spec.action_name) + '__'
```
dirk-thomas: Spelling: `prefix`.

jacobperron: Fixed in #28

PR #23, opened 2018-11-21
replace deprecated collections usage with collections.abc

dirk-thomas:
See https://docs.python.org/3/library/collections.html


dirk-thomas:

PR #26, opened 2019-01-04
Enforce ASCII characters on string and char fields

hidmic:
This pull request adds ASCII codification and de-codification steps whenever dealing with string and char message fields. 

Fixes https://github.com/ros2/ros2cli/issues/176 (which is actually one instance of an issue that is Python support wide).

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.c.em
+++ b/rosidl_generator_py/resource/_msg_support.c.em
@@ -201,13 +201,27 @@ lowercase_field_type = convert_camel_case_to_lower_case_underscore(field.type.ty
       }
 @[    if field.type.type == 'char']@
       assert(PyUnicode_Check(item));
-      @primitive_msg_type_to_c(field.type.type) tmp = PyUnicode_1BYTE_DATA(item)[0];
+      PyObject * encoded_item = PyUnicode_AsASCIIString(item);
+      if (!encoded_item) {
+        Py_DECREF(seq_field);
+        Py_DECREF(field);
+        return false;
+      }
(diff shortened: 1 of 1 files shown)


hidmic:
@dirk-thomas I believe you're the best person to review this. Not sure how this plays with (or if it's already fixed in) the refactored IDL code.

dirk-thomas:
The patch looks fine to me. Please run CI for the change and maybe add a test to check what happens when passing non-ASCII strings to it.

hidmic:
@dirk-thomas I think that writing a test for this is going to be somewhat involved. The conversions only take place within rclpy C extension code. I'd have to call publish on a message with a non-ascii string field to trigger a check when going from Python to C, and then take a message with a non-ascii string as sent by code that doesn't have such validations (e.g. a C++ node) to trigger a check when going from C to Python. Maybe a candidate for a `test_rclpy` package in `system_tests`? An easier alternative might be to write the test in C, using Python's C API.

dirk-thomas:
> I'd have to call publish on a message with a non-ascii string field to trigger a check when going from Python to C

That part sounds fairly straight forward?

> and then take a message with a non-ascii string as sent by code that doesn't have such validations (e.g. a C++ node) to trigger a check when going from C to Python. 

If the test only covers the Python-to-C case but not this one that would be sufficient for me.

hidmic:
> That part sounds fairly straight forward?

That's the easy part, yes. 

If you're OK with just testing Python-to-C, so am I.

hidmic:
CI's green now! We're good to go now.

PR #30, opened 2019-02-17
store types as tuple and return ordered dict

mikaelarguedas:
This is a follow-up of https://github.com/ros2/rosidl_python/pull/19
Required to fix https://github.com/ros2/ros2cli/issues/59

@dirk-thomas the returned keys right now are the exact names of the slots (with leading underscore), is it what you expected in your review of https://github.com/ros2/rosidl_python/pull/19#issuecomment-440058902 or should the leading underscore be stripped ?

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -116,27 +116,27 @@ class @(spec.base_type.type)(metaclass=Metaclass):
 @[end for]@
     ]
 
-    _fields_and_field_types = {
+    SLOT_TYPES = ((
 @[for field in spec.fields]@
 @[  if field.type.is_primitive_type() ]@
-        '@(field.name)': '@(field.type)',
+        '@(field.type)',
 @[  else ]@
 @[    if field.type.is_array ]@
 @[      if field.type.array_size ]@
 @[        if field.type.is_upper_bound ]@
-        '@(field.name)': '@(field.type.pkg_name)/@(field.type.type)[<=@(field.type.array_size)]',
+        '@(field.type.pkg_name)/@(field.type.type)[<=@(field.type.array_size)]',
 @[        else ]@
-        '@(field.name)': '@(field.type.pkg_name)/@(field.type.type)[@(field.type.array_size)]',
+        '@(field.type.pkg_name)/@(field.type.type)[@(field.type.array_size)]',
 @[        end if ]@
 @[      else ]@
-        '@(field.name)': '@(field.type.pkg_name)/@(field.type.type)[]',
+        '@(field.type.pkg_name)/@(field.type.type)[]',
 @[      end if ]@
 @[    else ]@
-        '@(field.name)': '@(field.type.pkg_name)/@(field.type.type)',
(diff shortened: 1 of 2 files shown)


rosidl_generator_py/resource/_msg.py.em
```diff
@@ -200,8 +199,9 @@ class @(spec.base_type.type)(metaclass=Metaclass):
         return True
 
     @@classmethod
-    def get_fields_and_field_types(cls):
-        return copy(cls._fields_and_field_types)
+    def get_slot_types(cls):
+        from collections import OrderedDict
```
dirk-thomas: Why not import the module at the top of the file?

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -116,27 +115,27 @@ class @(spec.base_type.type)(metaclass=Metaclass):
 @[end for]@
     ]
 
-    _fields_and_field_types = {
+    SLOT_TYPES = ((
```
dirk-thomas: Why two parenthesis here? One should be sufficient.

mikaelarguedas:
> I think it would make more sense to store an actual type object in the new class variable `SLOT_TYPES` - as suggested in the referenced comment:

Thanks for the clarification, I'll update to use these types instead.

mikaelarguedas:
@dirk-thomas Looking a bit more into this, it looks like this work will be conflicting / duplicating effort with https://github.com/ros2/rosidl_python/pull/24.

Would it be better to wait for #24 to land and build based on the state off master at that point?

--- 

If not, there is an initial working version at https://github.com/mikaelarguedas/rosidl_python/commit/0a88958a0374a231a9e0eb6c3b9c05b659a1a1c3 that I can push forward. A few questions remain:
How would consumer parse these types?
One option is to provide helper functions in a rosidl package to:
 - know if a basetype is a ROS primitive type or not (e.g. `string` is a ROS basic type but `String` is not a rosidl BasicType)
 - import a ROS module based on the `AbstractType` if it;s a complex type

Based on this the consumer could instantiate messages of that type and use them (use case from https://github.com/ros2/ros2cli/pull/197)

dirk-thomas:
> Looking a bit more into this, it looks like this work will be conflicting / duplicating effort with #24.

That will certainly be the case.

> Would it be better to wait for #24 to land and build based on the state off master at that point?

That would be great (to avoid the need for rebasing the patch). I hope that the referenced PR will get merged in the near future.

> How would consumer parse these types?

I am not sure I understand what you mean with "parse these types"? Since the API would return types (rather than strings) there shouldn't be a need to parse anything.

> * know if a basetype is a ROS primitive type or not (e.g. `string` is a ROS basic type but `String` is not a rosidl BasicType)

I don't think that distinction is relevant. At least in all the message generators patched for the IDL work I don't recall anywhere where a condition would say "isinstance(t, BaseType) or isinstance(t, BaseString)".

> import a ROS module based on the `AbstractType` if it;s a complex type

A function to import the module if the type is a `NamespaceType` might indeed be helpful.

mikaelarguedas:
Superseeded by https://github.com/ros2/rosidl_python/pull/33

PR #33, opened 2019-03-13
store types as tuple of abstract types

mikaelarguedas:
Replaces #30 now that #24 has merged

Required by ros2/ros2cli#207

Blocked by #34

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -69,7 +69,7 @@ for member in message.structure.members:
 @(import_statement)@
 @[        if import_statement not in import_statements]@
 @{import_statements.add(import_statement)}@
-  # noqa: E402@
+  # noqa: E402, I100@
 @[        end if]
 @[    end for]@
 @[end if]@
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -22,6 +22,7 @@ from rosidl_parser.definition import FLOATING_POINT_TYPES
 from rosidl_parser.definition import INTEGER_TYPES
 from rosidl_parser.definition import NamespacedType
 from rosidl_parser.definition import SIGNED_INTEGER_TYPES
+from rosidl_parser.definition import UnboundedSequence
 from rosidl_parser.definition import UNSIGNED_INTEGER_TYPES
 }@
 @#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
@@ -30,6 +31,9 @@ from rosidl_parser.definition import UNSIGNED_INTEGER_TYPES
 from collections import OrderedDict
 import numpy
 imports = OrderedDict()
+if message.structure.members:
+    imports.setdefault(
+        'import rosidl_parser.definition', [])  # used for SLOT_TYPES
 for member in message.structure.members:
     if (
         isinstance(member.type, AbstractNestedType) and
@@ -226,6 +230,39 @@ string@
 ',
 @[end for]@
     }
+
+    SLOT_TYPES = (
+@[for member in message.structure.members]@
+@{
+type_ = member.type
+if isinstance(type_, AbstractNestedType):
+    type_ = type_.value_type
+}@
+        @
+@[  if isinstance(member.type, AbstractNestedType)]@
+@(member.type.__class__.__module__).@(member.type.__class__.__name__)(@
+@[  end if]@
+@# the typename of the non-nested type or the nested value_type
+@(type_.__class__.__module__).@(type_.__class__.__name__)(@
+@[  if isinstance(type_, BasicType)]@
+'@(type_.typename)'@
+@[  elif isinstance(type_, AbstractGenericString) and type_.has_maximum_size()]@
+@(type_.maximum_size)@
+@[  elif isinstance(type_, NamespacedType)]@
+[@(', '.join("'%s'" % n for n in type_.namespaces))], '@(type_.name)'@
+@[  end if]@
+)@
+@[  if isinstance(member.type, Array)]@
+, @(member.type.size)@
+@[  elif isinstance(member.type, BoundedSequence)]@
+, @(member.type.maximum_size)@
+@[  end if]@
+@[  if isinstance(member.type, AbstractNestedType)]@
+)@
+@[  end if]@
+,  # noqa: E501
+@[end for]@
+    )
 
     def __init__(self, **kwargs):
         assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
--- a/rosidl_generator_py/test/test_interfaces.py
+++ b/rosidl_generator_py/test/test_interfaces.py
@@ -304,6 +304,63 @@ def test_out_of_range():
 
 def test_slot_attributes():
     a = Nested()
+    assert hasattr(a, 'get_fields_and_field_types')
+    assert hasattr(a, '__slots__')
+    nested_slot_types_dict = getattr(a, 'get_fields_and_field_types')()
+    nested_slots = getattr(a, '__slots__')
+    assert len(nested_slot_types_dict) == len(nested_slots)
+    expected_nested_slot_types_dict = {
+        'primitives': 'rosidl_generator_py/Primitives',
+        'two_primitives': 'rosidl_generator_py/Primitives[2]',
+        'up_to_three_primitives': 'sequence<rosidl_generator_py/Primitives, 3>',
+        'unbounded_primitives': 'sequence<rosidl_generator_py/Primitives>',
+    }
+    assert len(nested_slot_types_dict) == len(expected_nested_slot_types_dict)
+
+    for expected_field, expected_slot_type in expected_nested_slot_types_dict.items():
+        assert expected_field in nested_slot_types_dict.keys()
+        assert expected_slot_type == nested_slot_types_dict[expected_field]
+
+
+def test_string_slot_attributes():
+    b = StringArrays()
+    assert hasattr(b, 'get_fields_and_field_types')
+    assert hasattr(b, '__slots__')
+    string_slot_types_dict = getattr(b, 'get_fields_and_field_types')()
+    string_slots = getattr(b, '__slots__')
+    assert len(string_slot_types_dict) == len(string_slots)
+    expected_string_slot_types_dict = {
+        'ub_string_static_array_value': 'string<5>[3]',
+        'ub_string_ub_array_value': 'sequence<string<5>, 10>',
+        'ub_string_dynamic_array_value': 'sequence<string<5>>',
+        'string_dynamic_array_value': 'sequence<string>',
+        'string_static_array_value': 'string[3]',
+        'string_bounded_array_value': 'sequence<string, 10>',
+        'def_string_dynamic_array_value': 'sequence<string>',
+        'def_string_static_array_value': 'string[3]',
+        'def_string_bounded_array_value': 'sequence<string, 10>',
+        'def_various_quotes': 'sequence<string>',
+        'def_various_commas': 'sequence<string>',
+    }
+
+    assert len(string_slot_types_dict) == len(expected_string_slot_types_dict)
+
+    for expected_field, expected_slot_type in expected_string_slot_types_dict.items():
+        assert expected_field in string_slot_types_dict.keys()
+        assert expected_slot_type == string_slot_types_dict[expected_field]
+
+
+def test_modifying_slot_fields_and_types():
+    b = StringArrays()
+    assert hasattr(b, 'get_fields_and_field_types')
+    string_slot_types_dict = getattr(b, 'get_fields_and_field_types')()
+    string_slot_types_dict_len = len(string_slot_types_dict)
+    string_slot_types_dict[1] = 2
+    assert len(getattr(b, 'get_fields_and_field_types')()) == string_slot_types_dict_len
+
+
+def test_slot_types():
+    a = Nested()
     assert hasattr(a, 'SLOT_TYPES')
     assert hasattr(a, '__slots__')
     nested_slot_types = Nested.SLOT_TYPES
@@ -332,7 +389,7 @@ def test_slot_attributes():
     assert nested_slot_types[3].value_type.name == 'Primitives'
 
 
-def test_string_slot_attributes():
+def test_string_slot_types():
     b = StringArrays()
     assert hasattr(b, 'SLOT_TYPES')
     assert hasattr(b, '__slots__')
--- a/rosidl_generator_py/test/test_interfaces.py
+++ b/rosidl_generator_py/test/test_interfaces.py
@@ -27,9 +27,10 @@ from rosidl_generator_py.msg import WStrings
 
 from rosidl_parser.definition import Array
 from rosidl_parser.definition import BoundedSequence
+from rosidl_parser.definition import BoundedString
 from rosidl_parser.definition import NamespacedType
-from rosidl_parser.definition import String
 from rosidl_parser.definition import UnboundedSequence
+from rosidl_parser.definition import UnboundedString
 
 
 def test_strings():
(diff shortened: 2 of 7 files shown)


rosidl_generator_py/resource/_msg.py.em
```diff
@@ -135,45 +136,44 @@ class @(message.structure.type.name)(metaclass=Metaclass_@(message.structure.typ
...
 if isinstance(type_, NestedType):
     type_ = type_.basetype
 }@
-        '@(member.name)': '@
-@# the prefix for nested types
-@[  if isinstance(member.type, Sequence)]@
-sequence<@
+        @
+@[  if isinstance(member.type, BoundedSequence)]@
+rosidl_parser.definition.BoundedSequence(@
```
dirk-thomas: Instead of manually writing the name of each type I would suggest using `str(type(member.type))`.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -246,9 +246,8 @@ if isinstance(type_, NestedType):
         return True
 
     @@classmethod
-    def get_fields_and_field_types(cls):
-        from copy import copy
-        return copy(cls._fields_and_field_types)
+    def get_slot_types(cls):
```
dirk-thomas: This function should probably instead be implemented in the new runtime package proposed in #34.

mikaelarguedas: I'll wait for #34 to settle then.
Is there any other changes planned after #34 that would impact this ?

dirk-thomas: Hard to foresee the future :wink: I think this should be able to land right after #34 (even if other stuff comes up).

mikaelarguedas:
@dirk-thomas this has been rebased and modified to match changes from #34 and #35. 

In order to test that it fixes https://github.com/ros2/ros2cli/issues/59, it needs to be used along https://github.com/ros2/ros2cli/pull/209 (commit https://github.com/ros2/ros2cli/pull/209/commits/571d1ebdf575b1f2ae7f51a0a880ecc49a75d8ad)

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -182,45 +186,36 @@ class @(message.structure.type.name)(metaclass=Metaclass_@(message.structure.typ
...
 @[  elif isinstance(type_, BaseString)]@
-@
-@[    if isinstance(type_, WString)]@
-w@
-@[    end if]@
-string@
-@[    if type_.maximum_size is not None]@
-<@(type_.maximum_size)>@
-@[    end if]@
+'@(str(type_.maximum_size))')@
```
dirk-thomas: Why should the integer size be quoted?

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -182,45 +186,36 @@ class @(message.structure.type.name)(metaclass=Metaclass_@(message.structure.typ
...
 @[  elif isinstance(type_, NamespacedType)]@
-@('/'.join([type_.namespaces[0], type_.name]))@
+['@("', '".join(type_.namespaces))'], '@(type_.name)')@
```
dirk-thomas: This is wrong if `namespaces` is empty.

rosidl_runtime_py/rosidl_runtime_py/__init__.py
```diff
@@ -15,12 +15,16 @@
...
 from .set_message import set_message_fields
 
 
 __all__ = [
+    'import_message_from_namespaced_type',
     'message_to_csv',
     'message_to_ordereddict',
     'message_to_yaml',
     'set_message_fields',
+    'slot_types_dict_from_message',
```
dirk-thomas: Since functions should start with a verb how about `get_message_slot_types()`?

rosidl_runtime_py/rosidl_runtime_py/convert.py
```diff
@@ -156,3 +156,14 @@ def _convert_value(value, truncate_length=None):
...
         value = message_to_ordereddict(value, truncate_length=truncate_length)
     return value
+
+
+def slot_types_dict_from_message(msg: Any) -> OrderedDict:
+    """
+    Return an OrderedDict of the slot types of a message.
+
+    :param msg: The ROS message to convert.
+        This does not truncate the list of message fields.
```
dirk-thomas: I am not sure what this line is supposed to mean?

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -33,4 +37,13 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
         except TypeError:
             value = field_type()
             set_message_fields(value, field_value)
+        rosidl_type = slot_types_dict_from_message(msg)[field_name]
+        # Check if field is an array of ROS messages
+        if isinstance(field_type(), list):
```
dirk-thomas: Why not check the `rosidl_type` here?

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -271,56 +276,61 @@ def test_out_of_range():
...
-    string_slot_types_dict = getattr(b, 'get_fields_and_field_types')()
-    string_slot_types_dict_len = len(string_slot_types_dict)
-    string_slot_types_dict[1] = 2
-    assert len(getattr(b, 'get_fields_and_field_types')()) == string_slot_types_dict_len
+    assert len(string_slot_types) == len(string_slots)
+
+    assert isinstance(string_slot_types[0], Array)
+    assert isinstance(string_slot_types[0].basetype, String)
+    assert string_slot_types[0].size == 3
+    assert string_slot_types[0].basetype.maximum_size == '5'
```
dirk-thomas: Related to the comment above: this should be an `int`.

rosidl_generator_py/resource/_msg.py.em
mikaelarguedas: In which scenario would someone use a NamespacedType without namespace?

---

Is the expected syntax the following ?
```
NamespacedType([], 'myname')
```
instead of 
```
NamespacedType([''], 'myname')
```

---

I can fix it in here, wouldn't the other places in this file using this logic need to be updated as well ?
e.g. https://github.com/ros2/rosidl_python/blob/4fe21665960c3fefad5d46d3a6e559b39598018f/rosidl_generator_py/resource/_msg.py.em#L247

rosidl_generator_py/resource/_msg.py.em
mikaelarguedas: It shouldn't :+1: ; looks like a remnant from previous testings, removed in https://github.com/ros2/rosidl_python/pull/33/commits/808f12ff0a021bbc17272535ac84e0ecd4db757c. I also removed the string conversion as writing `None` explicitly is not needed either

rosidl_runtime_py/rosidl_runtime_py/__init__.py
mikaelarguedas: renamed in https://github.com/ros2/rosidl_python/pull/33/commits/4865fb48b24933c21fc7548771f71f7de56950a3

rosidl_runtime_py/rosidl_runtime_py/convert.py
mikaelarguedas: This was a copy-paste from other functions docblock, updated in https://github.com/ros2/rosidl_python/pull/33/commits/934aed2408f630d56b864626249e9a0afbed8e52

rosidl_runtime_py/rosidl_runtime_py/set_message.py
mikaelarguedas: It was to keep the condition simple.
Changed in https://github.com/ros2/rosidl_python/pull/33/commits/fbd5ecd01afafc42369041cee17b89798001f795

mikaelarguedas:
> A few other packages are currently using get_fields_and_field_types: rqt_py_common, rqt_plot, rqt_publisher, rqt_service_caller. Do you think it is possible to update those to use the new API too?

Yes it is possible, does this mean adding an explicit dependency on `rosidl_runtime_py` in all these packages ?

dirk-thomas:
> does this mean adding an explicit dependency on `rosidl_runtime_py` in all these packages ?

Only if they would use any symbols directly from that package. Otherwise no, each message package will already depend on `rosidl_runtime_py`.

rosidl_generator_py/resource/_msg.py.em
dirk-thomas: > In which scenario would someone use a NamespacedType without namespace?

There isn't a specific case where it is used atm but the code could be defensive and not rely on that assumption.

> `NamespacedType([], 'myname')`

Correct.

> wouldn't the other places in this file using this logic need to be updated as well ?

Probably, doesn't have to happen in this PR though. It's more a future proofing thing.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -33,4 +39,13 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
         except TypeError:
             value = field_type()
             set_message_fields(value, field_value)
+        rosidl_type = get_message_slot_types(msg)[field_name]
+        # Check if field is an array of ROS messages
+        if isinstance(rosidl_type, Array) or isinstance(rosidl_type, Sequence):
```
dirk-thomas: You can check for `isinstance(rosidl_type, NestedType)` which is the parent class of these two.

mikaelarguedas: `NestedType` is the one I needed :+1:  thanks!

mikaelarguedas:
> Only if they would use any symbols directly from that package. Otherwise no, each message package will already depend on rosidl_runtime_py.

Ok so either these packages use what is provided in the message itself (`__slots__` and `SLOT_TYPES`) or they depend explicitely on `rosidl_runtime_py` to be able to use the dictionary directly.

Do you have a preference for the rqt packages?

mikaelarguedas:
> > Do you have a preference for the rqt packages?

> I don't have a preference. Either way is fine with me.

Sounds good.
I don't know when I'll be able to allocate time for these, in case anyone else is interested in updating them in the meantime

rosidl_runtime_py/rosidl_runtime_py/import_message.py
```diff
@@ -0,0 +1,24 @@
...
+
+import importlib
+from typing import Any
+
+from rosidl_parser.definition import NamespacedType
+
+
+def import_message_from_namespaced_type(message_type: NamespacedType) -> Any:
+    module = importlib.import_module(
+            '.'.join(message_type.basetype.namespaces))
```
dirk-thomas: Nitpick: should be a single 4 space indentation.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -33,4 +38,13 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
...
             value = field_type()
             set_message_fields(value, field_value)
+        rosidl_type = get_message_slot_types(msg)[field_name]
+        # Check if field is an array of ROS messages
+        if isinstance(rosidl_type, NestedType):
+            if isinstance(rosidl_type.basetype, NamespacedType):
+                field_elem_type = import_message_from_namespaced_type(rosidl_type)
+                for n in range(len(value)):
+                    submsg = field_elem_type()
+                    set_message_fields(submsg, value[n])
```
dirk-thomas: `value[n]` looks weird here? Can you please extend the existing test of `set_message_fields` to cover this case of a nested namespaced type to make sure this works as expected.

mikaelarguedas: added test coverage in https://github.com/ros2/rosidl_python/pull/33/commits/ad65970d6bb4bff29a4d36fdc27845fb7d80f3f4

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -182,45 +186,36 @@ class @(message.structure.type.name)(metaclass=Metaclass_@(message.structure.typ
...
 type_ = member.type
 if isinstance(type_, NestedType):
     type_ = type_.basetype
 }@
-        '@(member.name)': '@
-@# the prefix for nested types
-@[  if isinstance(member.type, Sequence)]@
-sequence<@
+        @
+@[  if isinstance(member.type, Sequence) or isinstance(member.type, Array)]@
```
dirk-thomas: This could check for `NestedType` too.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -33,4 +38,13 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
         except TypeError:
             value = field_type()
```
dirk-thomas: I am surprised this works in cases where the field type is `numpy.ndarray` or `array.array`. Can you please also add a test with a static array (which would cover the `numpy` case).

mikaelarguedas: Yeah it surprised me as well that the code comparing to `list` didn't need to be updated after the numpy array PR..

Static array test case added in https://github.com/ros2/rosidl_python/pull/33/commits/712d4dafecad5fc5e416467a5f7d8a5f04b748dc

mikaelarguedas:
@dirk-thomas looking briefly at the impacted rqt packages:

In multiple instances it looks like the code is meant to be interacting with the user friendly string representing a message and not a Python type per se.
It looks like there is ongoing discussion on what that user friendly str identifying a type should be: https://github.com/ros2/design/pull/220

In some cases it's unclear if it is a design decision to rely on the `msg` type to be reflected in that dict.
e.g. in `rqt_py_common`, this API is used to [reconstruct a msg file from a class](https://github.com/ros-visualization/rqt/blob/5cf1e8f163cac2d0748c88976e71a40964d1297d/rqt_py_common/src/rqt_py_common/message_helpers.py#L333-L371). It looks to diverge significantly from the ROS 1 approach of providing the `msg` file on disk, and may require quite some custom code to be convert back from abstract type to msg format.
Is it a design decision to reconstruct `msg` files programmatically in the ROS 2 version?


Overall the amount of code + design in rqt to grasp and maybe rewrite will be significant, and I don't think I will have time to adapt it.

Is there a path forward for this PR without these rqt changes? I can reopen a PR not using the abstract types but the msg ones if this allows a fix for https://github.com/ros2/ros2cli/issues/59 to be merged

dirk-thomas:
I don't think this ticket has to consider what might come out of that design ticket. Until that is actually merged this PR can focus on the current state of master.

> Is it a design decision to reconstruct `msg` files programmatically in the ROS 2 version?

I am not entirely sure what you mean but I would answer with: no.

> Is there a path forward for this PR without these rqt changes?

As long as the patch removes currently available API which is used by rqt it can only land if it also updates the code using the API.

That being said the alternatives are:
* keep the old API in parallel to the newly introduced one
* don't change the API at all and work with what is there

dirk-thomas:
@mikaelarguedas Friendly ping.

mikaelarguedas:
So I looked at this again. as updating all the rqt packages is not something I will be able to do, I moved to a minimal patch not modifying message generation at all but just the runtime functions (a redo of https://github.com/ros2/ros2cli/pull/197/files).

I submitted a patch with the relevant tests commented out for review at https://github.com/ros2/rosidl_python/pull/52.

mikaelarguedas:
As per discussion here and in https://github.com/ros2/rosidl_python/pull/52.

This PR now does the following:

- Adds `SLOT_TYPES` to all messages
- Does not modify existing `_fields_and_field_types` or `get_fields_and_field_types` API
- Adds `import_message_from_namespaced_type` and `get_message_slot_types` to `rosidl_runtime_py`
- Update `set_message` to support Sequence and Arrays of ROS messages
- Add test to `rosidl_generator_py` and `rosidl_runtime_py` to cover new feature

dirk-thomas:
Thanks for the patch.

[Build Status](https://ci.ros2.org/job/ci_linux/6930/)

PR #34, opened 2019-03-15
Add rosidl_runtime_py package

jacobperron:
Migrated utility functions for working with Python ROS messages from the package `ros2topic`.

The intention is to reuse these functions in other CLI tools: `ros2service` and `ros2action`.

Replaces ros2/rclpy#285.

### DIFF
--- a/rosidl_runtime_py/package.xml
+++ b/rosidl_runtime_py/package.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rosidl_runtime_py</name>
+  <version>0.6.2</version>
+  <description>Runtime utilities for working with generated ROS interfaces in Python.</description>
+  <maintainer email="jacob@openrobotics.org">Jacob Perron</maintainer>
+  <license>Apache License 2.0</license>
+
+  <author email="dthomas@osrfoundation.org">Dirk Thomas</author>
+
+  <exec_depend>python3-numpy</exec_depend>
+  <exec_depend>python3-yaml</exec_depend>
+
+  <test_depend>ament_copyright</test_depend>
+  <test_depend>ament_flake8</test_depend>
+  <test_depend>ament_pep257</test_depend>
+  <test_depend>python3-pytest</test_depend>
+  <test_depend>test_msgs</test_depend>
+
+  <export>
+    <build_type>ament_python</build_type>
+  </export>
+</package>
--- a/rosidl_runtime_py/rosidl_runtime_py/convert.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/convert.py
@@ -1,0 +1,138 @@
+# Copyright 2016-2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import array
+from collections import OrderedDict
+import sys
+
+import numpy
+import yaml
+
+
+def register_yaml_representer():
+    # Register our custom representer for YAML output
+    yaml.add_representer(OrderedDict, represent_ordereddict)
+
+
+# Custom representer for getting clean YAML output that preserves the order in
+# an OrderedDict.
+# Inspired by:
+# http://stackoverflow.com/a/16782282/7169408
+def represent_ordereddict(dumper, data):
+    items = []
+    for k, v in data.items():
+        items.append((dumper.represent_data(k), dumper.represent_data(v)))
+    return yaml.nodes.MappingNode(u'tag:yaml.org,2002:map', items)
+
+
+def msg_to_yaml(args, msg):
+    return yaml.dump(
+        msg_to_ordereddict(
+            msg,
+            truncate_length=args.truncate_length if not args.full_length else None
+        ), width=sys.maxsize)
+
+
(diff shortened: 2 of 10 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6431/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2809/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5284/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6258/)

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6433/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2811/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5286/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6260/)

rosidl_runtime_py/package.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rosidl_runtime_py</name>
+  <version>0.6.2</version>
+  <description>Runtime utilities for working with generated ROS interfaces in Python.</description>
+  <maintainer email="scott@openrobotics.org">Scott K Logan</maintainer>
```
dirk-thomas: @jacobperron Is it intentionally that you create the package but put someone else as a maintainer into the manifest?

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
+    """
+    Set the fields of a ROS message.
+
+    :param msg: The ROS message to populate.
+    :param values: The values to set in the ROS message. The keys of the dictionary represent
+        fields of the message.
+    :raises SetFieldError: If there is an error setting a message field (e.g. the message does
```
dirk-thomas: Instead of always raising a custom exception (which prevents the calling code to distinguish the exceptions type using normal `except X` logic) why not raise specific exceptions for each case. For some even builtin exception types might be appropriate, e.g. `KeyError` when `values` contains a key which is not a field in the message.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -0,0 +1,57 @@
...
+        try:
+            field_type = type(getattr(msg, field_name))
+        except Exception as e:
+            raise SetFieldError(field_name, e)
+        try:
+            value = field_type(field_value)
+        except TypeError:
+            value = field_type()
+            try:
+                set_message_fields(value, field_value)
```
dirk-thomas: Calling this recursively in case of a `TypeError` seems weird to me. This should only happen when the field type is actually another message.

rosidl_runtime_py/rosidl_runtime_py/convert.py
```diff
@@ -0,0 +1,148 @@
...
+                    else:
+                        val += '...'
+            r = str(val)
+        else:
+            r = message_to_csv(val, truncate_length)
+        return r
+    result = ''
+    # We rely on __slots__ retaining the order of the fields in the .msg file.
+    for field_name in msg.__slots__:
+        value = getattr(msg, field_name, None)
```
dirk-thomas: Why should `. None` be passed? It seems like it is guaranteed that the `field_name` exists in the `msgs`?

rosidl_runtime_py/package.xml
jacobperron: No, this is a mistake.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
jacobperron: I've removed the custom exception in 4171c5a, simplifying the implementation.

rosidl_runtime_py/rosidl_runtime_py/convert.py
jacobperron: Not sure, this was from the original implementation. But I agree it looks like it's not necessary; removed in 4171c5a.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
jacobperron: From the original implementation. How can we check if the subtype is a message?

rosidl_runtime_py/package.xml
jacobperron: Fixed.

jacobperron:
I've gone back and grabbed the latest code from `ros2topic` and put the initial migration in its own commit before my changes so the diff should be more clear.

Conversion implementation taken from:

https://github.com/ros2/ros2cli/blob/978fc8807cb1f70b5f9c2b085e7b6fa74d191d0e/ros2topic/ros2topic/verb/echo.py

Set message implementation taken from:

https://github.com/ros2/ros2cli/blob/978fc8807cb1f70b5f9c2b085e7b6fa74d191d0e/ros2topic/ros2topic/api/__init__.py#L93-L119

jacobperron:
The subsequent replacement in ros2cli: https://github.com/ros2/ros2cli/pull/209

rosidl_runtime_py/rosidl_runtime_py/set_message.py
dirk-thomas: Each message class has a class method [get_fields_and_field_types](https://github.com/ros2/rosidl_python/blob/4f11d8734a5b78daf874a680359e813d42ca5fde/rosidl_generator_py/resource/_msg.py.em#L312) which provides the type of each slot.

But since it has worked this doesn't need to be changed in this PR (also that API is likely going to change soon, see #33).

jacobperron:
Thanks for the review!

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6554/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2893/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5371/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6352/)

PR #35, opened 2019-03-25
map Arrays to numpy.ndarray() and Sequences to array.array()

dirk-thomas:
Implements the latest version of the design article describing the type mapping for Python (including the latest change ros2/design#217).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6524/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2866/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5342/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6316/)

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -394,15 +394,15 @@ bound = 2**nbits
 @[  elif isinstance(type_, BasicType) and type_.type == 'octet']@
                 ((isinstance(value, bytes) or isinstance(value, ByteString)) and
                  len(value) == 1), \
-                "The '@(member.name)' field must of type 'bytes' or 'ByteString' with a length 1"
+                "The '@(member.name)' field must be of type 'bytes' or 'ByteString' with length 1"
 @[  elif isinstance(type_, BasicType) and type_.type == 'char']@
                 ((isinstance(value, str) or isinstance(value, UserString)) and
                  len(value) == 1 and ord(value) >= -128 and ord(value) < 128), \
-                "The '@(member.name)' field must of type 'str' or 'UserString' " \
-                'with a length 1 and the character ord() in [-128, 127]'
+                "The '@(member.name)' field must be of type 'str' or 'UserString' " \
+                'with length 1 and the character ord() in [-128, 127]'
 @[  elif isinstance(type_, BaseString)]@
                 isinstance(value, str), \
-                "The '@(member.name)' field must of type '@(get_python_type(type_))'"
+                "The '@(member.name)' field must be of type '@(get_python_type(type_))'"
 @[  elif isinstance(type_, BasicType) and type_.type in [
         'boolean',
         'float', 'double',
@@ -412,7 +412,7 @@ bound = 2**nbits
         'int64', 'uint64',
     ]]@
                 isinstance(value, @(get_python_type(type_))), \
-                "The '@(member.name)' field must of type '@(get_python_type(type_))'"
+                "The '@(member.name)' field must be of type '@(get_python_type(type_))'"
 @[    if type_.type.startswith('int')]@
 @{
 nbits = int(type_.type[3:])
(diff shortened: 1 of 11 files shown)


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6543/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2882/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5359/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6335/)

Ready for review.

dirk-thomas:
A great example of the performance improvement this change provides is to run the following two commands:
* `ros2 run image_tools cam2image -x 1280 -y 720`
* `ros2 topic echo /image`

Without this patch the Python CLI takes about 600-700ms to convert the incoming ROS message into a Python object. As a consequence only every 15th-20th frames is being shown and the command line tool uses 100% of one CPU core.

With the patch the conversion time is reduced by two orders of magnitude to about 7-8ms and the command line tool can print information for every image message without utilizing a full CPU core. Also the memory usage is reduced for the Python tool since an `array.array` uses only roughly the size of the underlaying C array (rather than a Python list with Python `int` items).

kervel:
Great, i think quite a few people would be helped by this (it would make the python API useable for us at least). See also https://github.com/ros2/ros2/issues/509 and various threads on discourse.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -18,6 +19,52 @@ from rosidl_parser.definition import Sequence
 from rosidl_parser.definition import String
 from rosidl_parser.definition import WString
 }@
+@#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
```
mjcarroll: Would you consider using a different delimiter here?  These look a lot like git conflict markers.

dirk-thomas: This kind of delimiter already exists in numerous em templates throughout the various generators. So for this PR I would rather stick with it. But a set of follow up PRs could replace them with whatever else is preferred.

mjcarroll: Ah, sorry, I guess I hadn't noticed it before.  It really jumped out at me this time.  I'm good with it staying, then.

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -209,11 +219,11 @@ def test_check_constraints():
         setattr(c, 'char_value', b'abc')
 
     c.up_to_three_int32_values = []
-    assert [] == c.up_to_three_int32_values
+    assert [] == c.up_to_three_int32_values.tolist()
```
hidmic: @dirk-thomas won't the list assignment right above break the test? Or are we using data descriptors or something?

dirk-thomas: Each message field has a property setter to ensure the right data type is being passed and the value range is valid. That setter also ensures that in this case an `array.array` is being stored (and not a `list`).

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -348,15 +432,15 @@ bound = 2**nbits
 @[  elif isinstance(type_, BasicType) and type_.type == 'octet']@
                 ((isinstance(value, bytes) or isinstance(value, ByteString)) and
                  len(value) == 1), \
-                "The '@(member.name)' field must of type 'bytes' or 'ByteString' with a length 1"
+                "The '@(member.name)' field must be of type 'bytes' or 'ByteString' with a length 1"
```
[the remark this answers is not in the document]
dirk-thomas: Applied in https://github.com/ros2/rosidl_python/pull/35/commits/43f0fe374fe62e81d97b8942759a12c254774516 to keep the DCO check happy.

dirk-thomas:
> pending green CI

@hidmic CI builds have already been run and are referenced in the comment above.

hidmic:
Oh, I see now that test failures are unrelated. LGTM then.

dirk-thomas:
Merging without squashing to keep the commits separate for easier history reading. Since these changes won't be backported the separated commits shouldn't be a problem.

dirk-thomas:
Addressed compiler warnings about unused variables in release mode in a follow up commit: 4f11d8734a5b78daf874a680359e813d42ca5fde.

PR #37, opened 2019-04-04
fix echo of numpy.number values

dirk-thomas:
Fixes ros2/ros2cli#213 which is a regression of #35.

Only Linux build since CI won't say much: [Build Status](https://ci.ros2.org/job/ci_linux/6599/)


mikaelarguedas:
Fixes the issue for me :+1:

PR #42, opened 2019-04-18
Fix quoted strings for new flake8-quote check.

clalancette:
Note that for generated code, we unconditionally disable the
flake8-quote check Q003 (Change outer quotes to avoid escaping
inner quotes).  For non-string types, this will have no effect.
For string types, it could be fairly expensive to scan strings
for single or double quotes.  The expense just doesn't seem
worth it and the resulting return value won't be exactly what
the user typed into their .msg file.

Connects to https://github.com/ros2/build_cop/issues/179

### DIFF
--- a/rosidl_generator_py/test/test_interfaces.py
+++ b/rosidl_generator_py/test/test_interfaces.py
@@ -114,9 +114,9 @@ def test_default_values():
     assert 'Hello world!' == Strings.DEF_STRING__DEFAULT
     assert 'Hello world!' == a.DEF_STRING__DEFAULT
 
-    assert 'Hello\'world!' == a.DEF_STRING2__DEFAULT
+    assert "Hello'world!" == a.DEF_STRING2__DEFAULT
     assert 'Hello"world!' == a.DEF_STRING3__DEFAULT
-    assert 'Hello\'world!' == a.DEF_STRING4__DEFAULT
+    assert "Hello'world!" == a.DEF_STRING4__DEFAULT
     assert 'Hello"world!' == a.DEF_STRING5__DEFAULT
     with pytest.raises(AttributeError):
         setattr(Strings, 'DEF_STRING__DEFAULT', 'bar')


clalancette:
CI run is in https://github.com/ros2/build_cop/issues/179#issuecomment-484509242

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -140,7 +140,7 @@ for member in message.structure.members:
 @[end for]@
 @[for member in message.structure.members]@
 @[  if member.has_annotation('default')]@
-            '@(member.name.upper())__DEFAULT': @(value_to_py(member.type, member.get_annotation_value('default')['value'])),
+            '@(member.name.upper())__DEFAULT': @(value_to_py(member.type, member.get_annotation_value('default')['value'])),  # noqa: Q003
```
dirk-thomas: Instead of applying the warning suppression to any kind of content I would suggest to update the `primitive_value_to_py()` function instead to use the proper quoting.

Same below.

clalancette: The problem is figuring out the proper quoting.  To do that, we actually have to walk the entire string and see whether there are any embedded single or double quotes, and then use the opposite.  We'll also need to handle escaped quotes properly.  In some cases, we won't be able to figure out one, so we'll still have to arbitrarily choose one.

It just doesn't seem worth it either for development time or for runtime.  What does the check really buy us?

dirk-thomas: That the generated code follows our code style guidelines.

clalancette: But that's just not a good enough reason, particularly for generated code, in my opinion.  All of this stuff has a cost in terms of time and complexity, and doing busywork like this just doesn't seem to be a good use of either.

dirk-thomas: I simply implemented the logic in a couple of minutes: see #43

clalancette:
All right, with #43 going in, we just need the changes to the test strings here, and we'll get rid of all of the flake8 warnings.  So this is ready for another review.

PR #45, opened 2019-04-19
simplify code using updated definition API

dirk-thomas:
Connect to ros2/rosidl#364.

### DIFF
--- a/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
+++ b/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
@@ -26,8 +26,11 @@ from rosidl_parser.definition import AbstractSequence
 from rosidl_parser.definition import AbstractString
 from rosidl_parser.definition import Array
 from rosidl_parser.definition import BasicType
+from rosidl_parser.definition import CHARACTER_TYPES
+from rosidl_parser.definition import FLOATING_POINT_TYPES
 from rosidl_parser.definition import IdlContent
 from rosidl_parser.definition import IdlLocator
+from rosidl_parser.definition import INTEGER_TYPES
 from rosidl_parser.definition import NamespacedType
 from rosidl_parser.parser import parse_idl_file
 
@@ -153,12 +156,7 @@ def primitive_value_to_py(type_, value):
     if type_.typename == 'boolean':
         return 'True' if value else 'False'
 
-    if type_.typename in (
-        'int8', 'uint8',
-        'int16', 'uint16',
-        'int32', 'uint32',
-        'int64', 'uint64',
-    ):
+    if type_.typename in INTEGER_TYPES:
         return str(value)
 
     if type_.typename == 'char':
@@ -167,7 +165,7 @@ def primitive_value_to_py(type_, value):
     if type_.typename == 'octet':
         return repr(bytes([value]))
 
-    if type_.typename in ('float', 'double'):
+    if type_.typename in FLOATING_POINT_TYPES:
         return '%s' % value
 
     assert False, "unknown primitive type '%s'" % type_.typename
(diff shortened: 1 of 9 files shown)


rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -180,12 +178,7 @@ def constant_value_to_py(type_, value):
         if type_.typename == 'bool':
```
hidmic: @dirk-thomas hmm shouldn't it be `'boolean'`? Also, here and everywhere else, why not use `BOOLEAN_TYPE`?

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -194,7 +187,7 @@ def constant_value_to_py(type_, value):
         if type_.typename == 'octet':
```
hidmic: @dirk-thomas same about `OCTET_TYPE`

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -180,12 +178,7 @@ def constant_value_to_py(type_, value):
...
-        if type_.typename in (
-            'int8', 'uint8',
-            'int16', 'uint16',
-            'int32', 'uint32',
-            'int64', 'uint64',
-        ):
+        if type_.typename in INTEGER_TYPES:
             return str(value)
 
         if type_.typename == 'char':
```
hidmic: @dirk-thomas same about an equivalent `CHAR_TYPE` (I know there's none, I just find it awkward to use the variable in some places and the raw string in others).

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
dirk-thomas: For a single typename it doesn't really make a difference. In cases where the constant represents a set of types it makes sense to use the constant in case the set changes later as well as for readability.

Since there are 100+ cases where single typenames are used as string literals in the various generators I would rather not touch it in this PR.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
dirk-thomas: See https://github.com/ros2/rosidl_python/pull/45#discussion_r277388399. Why should we create a constant for that name (`CHAR_TYPE = 'char'`)? What would be the advantage?

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
sloretz: The linter plugins can catch misspellings of `BOOLEAN_TYPE`, but not misspellings of `'bool'`. I don't think it needs to be done in this PR, just pointing out there is a benefit.

hidmic: I do see the improvement in readability, but if the sets were to ever change, having these variables won't do much against the other 100+ cases were the raw string is used (unless it's pure extension, of course). 

In any case, above `shouldn't it be 'boolean'` comment still applies, or otherwise I don't follow why `'boolean'` is used everywhere else when checking `BasicType.typename` attributes.

dirk-thomas: > In any case, above shouldn't it be 'boolean' comment still applies

Absolutely, we currently don't have any boolean constants. Adding one make the code fall through to the exception. I will create a follow PR to fix the behavior (since this doesn't aim to change any behavior).

dirk-thomas: > In any case, above shouldn't it be 'boolean' comment still applies

Absolutely, we currently don't have any boolean constants. Adding one make the code fall through to the exception. I created a follow PR to fix the behavior (since this doesn't aim to change any behavior): #46.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
dirk-thomas: Just copying a comment from above which applies here:

> The linter plugins can catch misspellings of `BOOLEAN_TYPE`, but not misspellings of `'bool'`.

PR #47, opened 2019-04-30
add WString support

dirk-thomas:
Connect to ros2/rosidl#352.

While waiting for ros2/rcl_interfaces#58 the patch (without tests) can already be reviewed.

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -376,7 +376,7 @@ if member.name in dict(inspect.getmembers(builtins)).keys():
             from collections.abc import Set
             from collections import UserList
             from collections import UserString
-@[  elif isinstance(type_, AbstractString) and type_.has_maximum_size()]@
+@[  elif isinstance(type_, AbstractGenericString) and type_.has_maximum_size()]@
             from collections import UserString
 @[  elif isinstance(type_, BasicType) and type_.typename == 'octet']@
             from collections.abc import ByteString
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -9,7 +9,6 @@ from rosidl_generator_py.generate_py_impl import value_to_py
 from rosidl_parser.definition import AbstractGenericString
 from rosidl_parser.definition import AbstractNestedType
 from rosidl_parser.definition import AbstractSequence
-from rosidl_parser.definition import AbstractString
 from rosidl_parser.definition import AbstractWString
 from rosidl_parser.definition import ACTION_FEEDBACK_SUFFIX
 from rosidl_parser.definition import ACTION_GOAL_SUFFIX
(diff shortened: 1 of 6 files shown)


rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -309,6 +319,31 @@ nested_type = '__'.join(type_.namespaced_name())
...
+      if (!encoded_item) {
+        Py_DECREF(seq_field);
+        Py_DECREF(field);
+        return false;
+      }
+      char * buffer;
+      Py_ssize_t length;
+      int rc = PyBytes_AsStringAndSize(encoded_item, &buffer, &length);
+      if (rc) {
+        Py_DECREF(seq_field);
```
jacobperron: We should decref `encoded_item` here.

rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -309,6 +319,31 @@ nested_type = '__'.join(type_.namespaced_name())
...
+        Py_DECREF(field);
+        return false;
+      }
+      // use offset of 2 to skip BOM mark
+      bool succeeded = rosidl_generator_c__U16String__assignn_from_char(&dest[i], buffer + 2, length - 2);
+      if (!succeeded) {
+        Py_DECREF(seq_field);
```
jacobperron: Decref `encoded_item` (or move decref to be before L334).

rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -370,6 +405,28 @@ nested_type = '__'.join(type_.namespaced_name())
...
+      Py_DECREF(field);
+      return false;
+    }
+    // use offset of 2 to skip BOM mark
+    bool succeeded = rosidl_generator_c__U16String__assignn_from_char(&ros_message->@(member.name), buffer + 2, length - 2);
+    if (!succeeded) {
+      Py_DECREF(field);
+      return false;
+    }
+    Py_DECREF(encoded_field);
```
jacobperron: We should ensure this is decref'd before or in the two preceding failure cases.

dirk-thomas: Good catch: see 03a891edf95220079d16febdf9ad3175647f0958.

rosidl_generator_py/resource/_msg_support.c.em
dirk-thomas: See https://github.com/ros2/rosidl_python/pull/47#discussion_r280238235

rosidl_generator_py/resource/_msg_support.c.em
dirk-thomas: See https://github.com/ros2/rosidl_python/pull/47#discussion_r280238235

rosidl_generator_py/resource/_msg_support.c.em
jacobperron: You moved one of them, but this one should still be moved.

dirk-thomas: And here it is: 1edd2365b6af17e46abdef5951aa154eb78fb7a8.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -376,7 +376,7 @@ if member.name in dict(inspect.getmembers(builtins)).keys():
             from collections.abc import Set
             from collections import UserList
             from collections import UserString
-@[  elif isinstance(type_, AbstractString) and type_.has_maximum_size()]@
+@[  elif isinstance(type_, AbstractGenericString) and type_.has_maximum_size()]@
```
jacobperron: What about the condition at L394 too?

dirk-thomas: Very good catch. Yes, that condition definitely needs update to check the length of WStrings too: 7fc6f0d2896faf8dba62d1edce3d0244ca2c115c

PR #49, opened 2019-05-01
[rosidl_runtime_py] Use new test interface definitions

jacobperron:
Connects to https://github.com/ros2/rcl_interfaces/issues/58


dirk-thomas:
Could it be that the function calls in `test_set_message.py` haven't been updated yet, e.g. https://github.com/ros2/rosidl_python/blob/230a6ab7ae407bc0ceb8ae26985f7b2ec01a0f09/rosidl_runtime_py/test/rosidl_runtime_py/test_set_message.py#L26

PR #52, opened 2019-05-02
Import message and support setting values of complex arrays/sequence

mikaelarguedas:
Version of https://github.com/ros2/rosidl_python/pull/33 not changing the API exposed by message packages.

This is a redo of https://github.com/ros2/ros2cli/pull/197 targetting this repository and updated to match what is returned by `get_fields_and_field_types`.

~~I left the tests commented out for now, they were passing about 2 days ago but since the message fixtures have been changed and the rosidl_runtime_py tests were not passing.
Submitting for review without updated tests. Hopefully we can update the tests shortly once the test_msgs situation stabilizes~~

Tests have been updated, ready for review

tested using command from https://github.com/ros2/ros2cli/issues/191#issue-403028625:
```
ros2 topic pub some_pose_array geometry_msgs/PoseArray "{header: {stamp: {sec: 1, nanosec: 2}, frame_id: some_frame}, poses: [{position: {x: 0, y: 1, z: 2}, orientation: {x: 3, y: 4, z: 5, w: 6}}]}"
```

### DIFF
--- a/rosidl_runtime_py/rosidl_runtime_py/import_message.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/import_message.py
@@ -1,0 +1,33 @@
+# Copyright 2017-2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib
+
+
+def import_message_type(topic_name: str, message_type: str):
+    # TODO(dirk-thomas) this logic should come from a rosidl related package
+    try:
+        package_name, message_name = message_type.split('/', 2)
+        if not package_name or not message_name:
+            raise ValueError()
+    except ValueError:
+        raise RuntimeError('The passed message type is invalid')
+
+    # TODO(sloretz) node API to get topic types should indicate if action or msg
+    middle_module = 'msg'
+    if topic_name.endswith('/_action/feedback'):
+        middle_module = 'action'
+
+    module = importlib.import_module(package_name + '.' + middle_module)
+    return getattr(module, message_name)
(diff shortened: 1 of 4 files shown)


rosidl_runtime_py/rosidl_runtime_py/import_message.py
```diff
@@ -0,0 +1,33 @@
...
+        package_name, message_name = message_type.split('/', 2)
+        if not package_name or not message_name:
+            raise ValueError()
+    except ValueError:
+        raise RuntimeError('The passed message type is invalid')
+
+    # TODO(sloretz) node API to get topic types should indicate if action or msg
+    middle_module = 'msg'
+    if topic_name.endswith('/_action/feedback'):
+        middle_module = 'action'
```
dirk-thomas: Once this problem has been addressed the API would not need to take `topic_name` anymore which would break users.  Therefore I don't think this API should added as-is.

mikaelarguedas: I would expect in the future the import function to be something like https://github.com/ros2/rosidl_python/pull/33/files#diff-2283820db2117771f9254b0ff5ac87f0 using the abstract types.


What would be an acceptable API in the meantime? a keyword argument for the `topic_name` that would default to an empty string? (note that it would still break API the day that keyword argument is removed).

Or would you rather keep the original function in ros2topic and make a duplicated function here without this argument at all ?

dirk-thomas: I would rather not add API here which is known that it needs to change as soon as something else gets fixed. So for now the API can stay were it is and once a stable API can be created it can land in the `rosidl_runtime_py` package.

Also I just created #54 which adds the `SLOT_TYPES` to each message. Maybe that is sufficient for what you want to achieve?

mikaelarguedas: Thanks, just to clarify my understanding:
The suggestion is to use the `SLOT_TYPES` from #54 
introduce here the new `import_message_from_namespaced_type` from #33
and modify set_message to use these 2 new bits in order to fill Nested ROS messages

Is that right ?

dirk-thomas: Yeah, that should work.

mikaelarguedas: Alright I integrated the changes from #54 and updated https://github.com/ros2/rosidl_python/pull/33 accordingly.

mikaelarguedas:
Superseeded by #33

PR #55, opened 2019-05-08
fix conversion from C to Python in case a sequence has default values

dirk-thomas:
Since we currently don't have a way to construct a Python message class without its default values we have to remove them before populating the `array.array` fields using `frombytes`.

Related to ros2/test_interface_files#5.

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.c.em
+++ b/rosidl_generator_py/resource/_msg_support.c.em
@@ -517,24 +517,44 @@ if isinstance(type_, AbstractNestedType):
     memcpy(dst, src, @(member.type.size) * sizeof(@primitive_msg_type_to_c(member.type.value_type)));
     Py_DECREF(field);
 @[    elif isinstance(member.type, AbstractSequence)]@
+    field = PyObject_GetAttrString(_pymessage, "@(member.name)");
+    if (!field) {
+      return NULL;
+    }
+    assert(field->ob_type != NULL);
+    assert(field->ob_type->tp_name != NULL);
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Passed CI, see https://github.com/ros2/test_interface_files/pull/5#issuecomment-490680355

rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -517,24 +517,44 @@ if isinstance(type_, AbstractNestedType):
...
+    PyObject * itemsize_attr = PyObject_GetAttrString(field, "itemsize");
+    assert(itemsize_attr != NULL);
+    size_t itemsize = PyLong_AsSize_t(itemsize_attr);
+    Py_DECREF(itemsize_attr);
+    if (itemsize != sizeof(@primitive_msg_type_to_c(member.type.value_type))) {
+      PyErr_SetString(PyExc_RuntimeError, "itemsize doesn't match expectation");
+      Py_DECREF(field);
+      return NULL;
+    }
+    // clear the array, poor approach to remove potential default values
```
jacobperron: We could also create a new array object with the same typecode. But I'm not sure if this would be any prettier, or if we still need to worry about decref'ing the individual elements.

In any case, this looks okay to me with green CI.

dirk-thomas: I would be happy to do so but couldn't figure out how to instantiate an `array.array` from the C code when I wrote this originally :disappointed:

PR #57, opened 2019-05-15
encode/decode strings with UTF-8

dirk-thomas:
This patch rolls back the changes from #26 and encodes / decodes (non W-)strings as UTF-8.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7091/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3317/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5813/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6913/)

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.c.em
+++ b/rosidl_generator_py/resource/_msg_support.c.em
@@ -321,7 +321,7 @@ nested_type = '__'.join(type_.namespaced_name())
       Py_DECREF(encoded_item);
 @[    elif isinstance(member.type.value_type, AbstractWString)]@
       assert(PyUnicode_Check(item));
-      // the returned string starts with a BOM mark
+      // the returned string starts with a BOM mark and uses native byte order
       PyObject * encoded_item = PyUnicode_AsUTF16String(item);
       if (!encoded_item) {
         Py_DECREF(seq_field);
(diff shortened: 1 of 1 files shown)


hidmic:
> This patch rolls back the changes from #26 

Does it? Also, I wonder, will this play along with other non-Python nodes? Encoded UTF-8 isn't ASCII nor any stream of bytes is UTF-8 decodable. What's the premise here?

dirk-thomas:
> Does it?

It doesn't revert the exact change but rolls back the semantic change of enforcing ASCII.

> will this play along with other non-Python nodes?

Yes, the referenced test cases cover C++ as well as Python and the tests in `test_communication` ensure that the information is correctly passed between languages.

> Encoded UTF-8 isn't ASCII nor any stream of bytes is UTF-8 decodable.

We follow the X-Types spec to encode (non W-)strings with UTF-8. If you only use ASCII characters nothing changes here since all ASCII characters have the same representation when encoded with UTF-8. If you do use non-ASCII characters they are being encoded, the wire format contains UTF-8 encoded bytes, and the receiver side decodes the string.

If you have an external entity which publishes strings with data which can't be UTF-8 decoded (an violate the X-Types spec) our receiving nodes will bail.

hidmic:
> We follow the X-Types spec

Oh, that's it. Perfect. LGTM too!

PR #59, opened 2019-05-30
Fix the setting of array variables.

clalancette:
If the type of the field is array.array, then a simple

  x = array.array(mylist)

doesn't work; the typecode *must* be specified.  Special-case
arrays here so that we do:

  x = array.array(typecode, list)

This should fix https://github.com/ros2/ros2cli/issues/270

### DIFF
--- a/rosidl_runtime_py/rosidl_runtime_py/set_message.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/set_message.py
@@ -34,15 +34,18 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
     :raises ValueError: If a message value does not match its field type.
     """
     for field_name, field_value in values.items():
-        field_type = type(getattr(msg, field_name))
+        field = getattr(msg, field_name)
+        field_type = type(field)
         try:
-            if field_type == array.array:
-                field_typecode = getattr(msg, field_name).typecode
-                value = field_type(field_typecode, field_value)
+            if field_type is array.array:
+                value = field_type(field.typecode, field_value)
             else:
                 value = field_type(field_value)
         except TypeError:
-            value = field_type()
+            if field_type is array.array:
+                value = field_type(field.typecode)
+            else:
+                value = field_type()
             set_message_fields(value, field_value)
         rosidl_type = get_message_slot_types(msg)[field_name]
         # Check if field is an array of ROS messages
--- a/rosidl_runtime_py/rosidl_runtime_py/set_message.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/set_message.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import array
+
 from typing import Any
 from typing import Dict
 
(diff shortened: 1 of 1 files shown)


rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -34,7 +36,11 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
...
         field_type = type(getattr(msg, field_name))
         try:
-            value = field_type(field_value)
+            if field_type == array.array:
+                field_typecode = getattr(msg, field_name).typecode
+                value = field_type(field_typecode, field_value)
+            else:
+                value = field_type(field_value)
         except TypeError:
             value = field_type()
```
clalancette: If we ever hit a `TypeError` when the `field_type` is `array.array`, then we'll throw a different exception here since you can't instantiate an `array.array` without a typecode.  However, I couldn't see how that would happen, so I didn't make any changes here.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -34,7 +36,11 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
     for field_name, field_value in values.items():
         field_type = type(getattr(msg, field_name))
         try:
-            value = field_type(field_value)
+            if field_type == array.array:
```
hidmic: @clalancette nit: 
```suggestion
            if field_type is array.array:
```

dirk-thomas:
Can you please check if sequences (with a dynamic size) are also affect and need a similar fix.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
dirk-thomas: Please apply the same conditional logic in the exception case to construct and empty array.

clalancette:
> Can you please check if sequences (with a dynamic size) are also affect and need a similar fix.

Good call, will check on that next.

clalancette:
> Good call, will check on that next.

As far as I can tell, the sequence stuff works (at least, basically).  I ran:

```
ros2 topic pub -r 5 /foo std_msgs/Int32MultiArray '{layout: {dim: [{label: 'foo', size: 5, stride: 4}], data_offset: 12}, data: [1, 5]}'
```

Which seemed to work fine.  In retrospect, that's because Sequences (lists and tuples, at least) can be instantiated with just `val = list(list)` or whatever.  I'll point out that ranges (which are technically a Sequence) don't have this property and need to be instantiated with at least one value.  Is that something we  worry about?

There is also a different bug in there, in that sending the wrong type for the data gives an opaque error message, but I'll ticket that separately.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
clalancette: Fixed in a29f80f

rosidl_runtime_py/rosidl_runtime_py/set_message.py
clalancette: Fixed in a29f80f

clalancette:
https://github.com/ros2/rosidl_python/issues/61 is the follow-up.  It's in the same area, but I'm running out of time today.  Since it's a cosmetic issue, I think we can punt to after Dashing.  CI up to rosidl_runtime_py:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7274/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3452/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5952/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7088/)

dirk-thomas:
See #62 for a revised patch which doesn't result in #61.

PR #60, opened 2019-05-30
Make the message __repr__ for python look nicer.

clalancette:
Before this patch, publishing a message with "ros2 topic pub"
would print something like:

publishing #5: my_msgs.msg.my_msg(axes=array('f', [0.0]))

While that is OK, it is kind of ugly.  This patch hides the
typecode and the "array" so that the output looks like:

publishing #5: my_msgs.msg.my_msg(axes=[0.0])

Which is also how it looked in Crystal.

Note that this is currently targeting the `fix-array-publish` branch; we'll need to retarget once that one is merged.  Also note that this is kind of an RFC; while it is quite ugly, it is a cosmetic issue, and touching the generated python code may be too risky for Dashing at this point.  I'll run CI presently, but opinions welcome.

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -334,23 +334,19 @@ if isinstance(type_, AbstractNestedType):
         typename = self.__class__.__module__.split('.')
         typename.pop()
         typename.append(self.__class__.__name__)
-@[if 'import array' in imports]@
         args = []
-        for s in self.__slots__:
+        for s,t in zip(self.__slots__, self.SLOT_TYPES):
             field = getattr(self, s, None)
-            t = repr(field)
-            if type(field) is array.array:
+            fieldstr = repr(field)
+            if isinstance(t, rosidl_parser.definition.AbstractSequence):
                 # Calling `.tolist()` on a large array could be memory
                 # expensive, so instead we first repr() it, then strip off the
                 # "array('i')" or "array('i', [1])".
-                if t[9:11] == ', ':
-                    t = t[11:-1]
-                elif len(t) == 10:
-                    t = '[]'
-            args.append(s[1:] + '=' + t)
-@[else]@
-        args = [s[1:] + '=' + repr(getattr(self, s, None)) for s in self.__slots__]
-@[end if]@
+                if fieldstr[9:11] == ', ':
+                    fieldstr = fieldstr[11:-1]
+                elif len(fieldstr) == 10:
+                    fieldstr = '[]'
+            args.append(s[1:] + '=' + fieldstr)
         return '%s(%s)' % ('.'.join(typename), ', '.join(args))
 
     def __eq__(self, other):
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -335,7 +335,7 @@ if isinstance(type_, AbstractNestedType):
         typename.pop()
         typename.append(self.__class__.__name__)
         args = []
-        for s,t in zip(self.__slots__, self.SLOT_TYPES):
+        for s, t in zip(self.__slots__, self.SLOT_TYPES):
             field = getattr(self, s, None)
             fieldstr = repr(field)
             if isinstance(t, rosidl_parser.definition.AbstractSequence):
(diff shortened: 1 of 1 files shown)


rosidl_generator_py/resource/_msg.py.em
```diff
@@ -334,7 +334,18 @@ if isinstance(type_, AbstractNestedType):
         typename = self.__class__.__module__.split('.')
         typename.pop()
         typename.append(self.__class__.__name__)
+@[if 'import array' in imports]@
```
dirk-thomas: Instead of using this runtime condition (which magically changes the logic for all fields as soon as an array field is added) use the types stored for each slot in `SLOT_TYPES`.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -334,7 +334,18 @@ if isinstance(type_, AbstractNestedType):
...
+        args = []
+        for s in self.__slots__:
+            field = getattr(self, s, None)
+            if type(field) == array.array:
+                t = repr(field.tolist())
```
dirk-thomas: Since this operation is fairly expensive (in terms of memory usage) I would suggest to keep using `repr(field)` and then process the string to remove the `array(typecode, )` part.

clalancette: I did this in 079d988.

rosidl_generator_py/resource/_msg.py.em
clalancette: I see, good point.  I've done this in 33e0ff68eac0e27030488b9b1e8683cc1ea063bc

clalancette:
Since this touches all of the generated Python code, I'm going to do a full CI run on Linux: * Linux [Build Status](http://ci.ros2.org/job/ci_linux/7276/)

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -334,7 +334,19 @@ if isinstance(type_, AbstractNestedType):
         typename = self.__class__.__module__.split('.')
         typename.pop()
         typename.append(self.__class__.__name__)
-        args = [s[1:] + '=' + repr(getattr(self, s, None)) for s in self.__slots__]
+        args = []
+        for s, t in zip(self.__slots__, self.SLOT_TYPES):
+            field = getattr(self, s, None)
+            fieldstr = repr(field)
+            if isinstance(t, rosidl_parser.definition.AbstractSequence):
```
clalancette: Ug, I don't think this will work for other `AbstractSequence` types; I need to be more selective.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -334,7 +334,19 @@ if isinstance(type_, AbstractNestedType):
         typename = self.__class__.__module__.split('.')
         typename.pop()
         typename.append(self.__class__.__name__)
-        args = [s[1:] + '=' + repr(getattr(self, s, None)) for s in self.__slots__]
+        args = []
+        for s, t in zip(self.__slots__, self.SLOT_TYPES):
+            field = getattr(self, s, None)
```
dirk-thomas: Nitpick: no need for `, None`.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -334,7 +334,19 @@ if isinstance(type_, AbstractNestedType):
...
+            fieldstr = repr(field)
+            if isinstance(t, rosidl_parser.definition.AbstractSequence):
+                # Calling `.tolist()` on a large array could be memory
+                # expensive, so instead we first repr() it, then strip off the
+                # "array('i')" or "array('i', [1])".
+                if fieldstr[9:11] == ', ':
+                    fieldstr = fieldstr[11:-1]
+                elif len(fieldstr) == 10:
```
dirk-thomas: This condition can be more easily expressed as `len(field) == 0`.

jacobperron:
I guess we may only see the array prefix with arrays of basic types.

dirk-thomas:
> Why is it that only some messages are affected?

`array.array` is only used for numeric types: http://design.ros2.org/articles/idl_interface_definition.html#type-mapping

PR #62, opened 2019-05-30
fix logic around populating arrays

### LINKED ISSUE #61: Trying to publish an invalid array type gives opaque error message
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Source
- Version or commit hash:
  - c9a471158dcc4ad861e2069261e43e7940f673ad with fix from #59 
- DDS implementation:
  - Opensplice
  - rclpy

#### Steps to reproduce issue

```
ros2 topic pub -r 5 /foo std_msgs/Int32MultiArray '{layout: {dim: [{label: 'foo', size: 5, stride: 4}], data_offset: 12}, data: [1, 5.0]}'
```

#### Expected behavior

An error message telling me that I gave an invalid value to the list of data (this is an Int32 array, so the 5.0 is illegal).

#### Actual behavior

```
Failed to populate field: 'list' object has no attribute 'items'
```

#### Additional information

This happens because the code at https://github.com/ros2/rosidl_python/pull/59/files#diff-5730f1f74a20f50bf6a7e53b26d7f034R41 throws a `TypeError` exception when trying to stuff a double into an integer `array.array`.  When that happens, we instantiate an empty `array.array`, then try to recursively call `set_message_fields` with the `field_value`.  However, at this point `field_value` is a list of `[1, 5.0]`, so the recursive call tries to call `.items()` on the list and fails.

dirk-thomas:
Fixes #61 which is a consequence direct consequence of #59.

For array fields there is no need to try/except the construction. If it fails the exception should be raised as is.

This makes both examples work:

* `ros2 topic pub -r 5 /joy sensor_msgs/Joy "{header:{stamp:{sec: 0, nanosec: 0}, frame_id: 'frame'}, axes:[0.0], buttons:[0]}"`
* `ros2 topic pub -r 5 /foo std_msgs/Int32MultiArray '{layout: {dim: [{label: 'foo', size: 5, stride: 4}], data_offset: 12}, data: [1, 5.0]}'`

### DIFF
--- a/rosidl_runtime_py/rosidl_runtime_py/set_message.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/set_message.py
@@ -36,17 +36,14 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
     for field_name, field_value in values.items():
         field = getattr(msg, field_name)
         field_type = type(field)
-        try:
-            if field_type is array.array:
-                value = field_type(field.typecode, field_value)
-            else:
+        if field_type is array.array:
+            value = field_type(field.typecode, field_value)
+        else:
+            try:
                 value = field_type(field_value)
-        except TypeError:
-            if field_type is array.array:
-                value = field_type(field.typecode)
-            else:
+            except TypeError:
                 value = field_type()
-            set_message_fields(value, field_value)
+                set_message_fields(value, field_value)
         rosidl_type = get_message_slot_types(msg)[field_name]
         # Check if field is an array of ROS messages
         if isinstance(rosidl_type, AbstractNestedType):


dirk-thomas:
CI up to `rosidl_runtime_py`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7275/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3453/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5953/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7089/)

PR #63, opened 2019-05-31
Ensure the contents of the field are an array.

clalancette:
It turns out that while all AbstractSequences are *defined*
to be arrays, we allow users to assign other sequences (like
lists) to those fields.  Make sure that the field starts with
'array(' before trying to strip it off.

This should fix https://github.com/ros2/build_cop/issues/212

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -339,13 +339,16 @@ if isinstance(type_, AbstractNestedType):
             field = getattr(self, s)
             fieldstr = repr(field)
             if isinstance(t, rosidl_parser.definition.AbstractSequence):
-                if len(field) == 0:
-                    fieldstr = '[]'
-                else:
-                    assert fieldstr.startswith('array(')
-                    prefix = "array('X', "
-                    suffix = ')'
-                    fieldstr = fieldstr[len(prefix):-len(suffix)]
+                # It is possible that the user assigned a non-array type to
+                # the AbstractSequence field, so only remove 'array(' if it
+                # is in the string.
+                if fieldstr.startswith('array('):
+                    if len(field) == 0:
+                        fieldstr = '[]'
+                    else:
+                        prefix = "array('X', "
+                        suffix = ')'
+                        fieldstr = fieldstr[len(prefix):-len(suffix)]
             args.append(s[1:] + '=' + fieldstr)
         return '%s(%s)' % ('.'.join(typename), ', '.join(args))
 
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -342,8 +342,8 @@ if isinstance(type_, AbstractNestedType):
             # in them, and "normal" sequences for everything else.  If it is
             # a type that we store in an array, strip off the 'array' portion.
             if (
-                isinstance(t, rosidl_parser.definition.AbstractSequence) and \
-                isinstance(t.value_type, rosidl_parser.definition.BasicType) and \
+                isinstance(t, rosidl_parser.definition.AbstractSequence) and
+                isinstance(t.value_type, rosidl_parser.definition.BasicType) and
                 t.value_type.typename in @([*SPECIAL_NESTED_BASIC_TYPES])
             ):
                 if len(field) == 0:
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -339,16 +339,15 @@ if isinstance(type_, AbstractNestedType):
             field = getattr(self, s)
             fieldstr = repr(field)
             if isinstance(t, rosidl_parser.definition.AbstractSequence):
-                # It is possible that the user assigned a non-array type to
-                # the AbstractSequence field, so only remove 'array(' if it
-                # is in the string.
-                if fieldstr.startswith('array('):
-                    if len(field) == 0:
-                        fieldstr = '[]'
-                    else:
-                        prefix = "array('X', "
-                        suffix = ')'
-                        fieldstr = fieldstr[len(prefix):-len(suffix)]
+                if len(field) == 0:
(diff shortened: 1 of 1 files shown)


clalancette:
Here's a full CI run:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7279/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3455/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5955/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7091/)

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -339,13 +339,16 @@ if isinstance(type_, AbstractNestedType):
...
-                    fieldstr = '[]'
-                else:
-                    assert fieldstr.startswith('array(')
-                    prefix = "array('X', "
-                    suffix = ')'
-                    fieldstr = fieldstr[len(prefix):-len(suffix)]
+                # It is possible that the user assigned a non-array type to
+                # the AbstractSequence field, so only remove 'array(' if it
+                # is in the string.
+                if fieldstr.startswith('array('):
```
dirk-thomas: Please move the check into the `else` block since there is no reason for it on the `if` block.

dirk-thomas:
When the user assigns a list to an array field I would expect the setter function to ensure that the actually stored value is an `array.array`: see https://github.com/ros2/rosidl_python/blob/f78c207669a71cbea537684d654d96e592c5c083/rosidl_generator_py/resource/_msg.py.em#L520-L524

So I think before making this change we need to investigate why that is not the case (and maybe address that instead).

clalancette:
> So I think before making this change we need to investigate why that is not the case (and maybe address that instead).

Yeah, that occurred to me too.  I put this up so we could get it in quickly, but since this isn't a rush for release now we should probably investigate that.

rosidl_generator_py/resource/_msg.py.em
clalancette: Done in 810ac05

clalancette:
> So I think before making this change we need to investigate why that is not the case (and maybe address that instead).

I see where the discrepancy is.  The problem shows up for types that are not in the [SPECIAL_NESTED_BASIC_TYPES](https://github.com/ros2/rosidl_python/blob/master/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py#L37).  If the message has a Sequence of booleans, for instance, there is no way to directly map them to an [array](https://docs.python.org/3.7/library/array.html).  So the generated setter only checks if the new value is a valid Sequence, and does a direct assignment.  While we could potentially think about coercion to an `array` for some of the primitive types, this wouldn't work in general for Sequences of non-primitives.

I think the fallout from this is that we do need to handle regular Sequences along with `array` types, and hence this patch still makes sense.  Thoughts?

dirk-thomas:
Based on your argument I don't think the current patch is right. I would rather change the condition `isinstance(t, rosidl_parser.definition.AbstractSequence)` to also check for `SPECIAL_NESTED_BASIC_TYPES`.

clalancette:
> `` to also check for `SPECIAL_NESTED_BASIC_TYPES`.

The problem with that is that SPECIAL_NESTED_BASIC_TYPES comes from https://github.com/ros2/rosidl_python/blob/master/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py#L37 , and we don't currently import rosidl_generator_py at runtime for the generated messages.  Our options seem to be:

1.  Import rosidl_generator_py at runtime
     1.  Do it at the top of the file
     1.  Do it just in `repr` (slightly better performance if `repr` is never called)
1.  Move SPECIAL_NESTED_BASIC_TYPES to rosidl_parser.definition (I'm not sure what additional changes would be required with that)
1.  Keep with what this patch does and use an indirect check (presence of 'array(') to determine this

I'm not really hugely in favor of importing `rosidl_generator_py` at runtime, mostly because we slow down all message code with an additional import.  However, I could be convinced otherwise.

dirk-thomas:
How about 4. use `SPECIAL_NESTED_BASIC_TYPES` in the template and expand the actual values into the generated code?

clalancette:
> How about 4. use `SPECIAL_NESTED_BASIC_TYPES` in the template and expand the actual values into the generated code?

Done in 0e8e345.  New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7284/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3460/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5960/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7098/)

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -338,7 +338,12 @@ if isinstance(type_, AbstractNestedType):
         for s, t in zip(self.__slots__, self.SLOT_TYPES):
             field = getattr(self, s)
             fieldstr = repr(field)
-            if isinstance(t, rosidl_parser.definition.AbstractSequence):
+            # We use Python array type for fields that can be directly stored
+            # in them, and "normal" sequences for everything else.  If it is
+            # a type that we store in an array, strip off the 'array' portion.
+            if isinstance(t, rosidl_parser.definition.AbstractSequence) and \
+               isinstance(t.value_type, rosidl_parser.definition.BasicType) and \
+               t.value_type.typename in @([*SPECIAL_NESTED_BASIC_TYPES]):
```
dirk-thomas: Nitpick: the multi line condition isn't distinguishable from the indented block. Please use the following style instead:

```
if (
    ... and
    ...
):
    ...
```

clalancette: Done in f7b18f6e065cea676e8a13d1a3a83043094f307a

clalancette:
CI is green (Windows test failures are unrelated), so I'm going to merge.

PR #73, opened 2019-07-18
remove the padding member from structs which should be empty

dirk-thomas:
IDL and some programming languages like C require a struct to have at least one member. Python though doesn't have that restriction and can define a class without any slots.

Therefore this patch removes the member `structure_needs_at_least_one_member` only used to "pad" empty structures from the Python data type. As a consequence printing the message won't show the member as well as the user won't see a getter and setter for the member anymore.

In the conversion from Python to C the member variable present in C is just set to `0` in order to not be uninitialized. In the other direction the value in the C struct is just ignored and not transferred to the Python object.

Fixes ros2/ros2cli#297. Requires ros2/rosidl#389.

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -18,6 +18,7 @@ from rosidl_parser.definition import BasicType
 from rosidl_parser.definition import BOOLEAN_TYPE
 from rosidl_parser.definition import BoundedSequence
 from rosidl_parser.definition import CHARACTER_TYPES
+from rosidl_parser.definition import EMPTY_STRUCTURE_REQUIRED_MEMBER_NAME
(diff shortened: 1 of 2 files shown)


sloretz:
I think this spot at line 379 in `__eq__` might need to be skipped

https://github.com/ros2/rosidl_python/blob/000cf6b5500621b5827c489b5e340e570238cc22/rosidl_generator_py/resource/_msg.py.em#L372-L382

dirk-thomas:
> I think this spot at line 379 in __eq__ might need to be skipped

Very good point. Fixes in d97cc46ee9892fa6b36e65970946de3b4c6cd3d0. I am curious if the running CI builds will complain about it :thinking:

dirk-thomas:
CI build including the second commit: [Build Status](https://ci.ros2.org/job/ci_linux/7613/) (same test failures as with master)

PR #74, opened 2019-07-20
Make import_message_from_namespaced_type operate on NamespacedType

### LINKED ISSUE #70: Erroneous mypy specifications
Edit:
I've been using `import_message_from_namespaced_type` wrong due to misleading erroneous mypy specifications
```python
def import_message_from_namespaced_type(message_type: NamespacedType)
```

Leaving this issue open until the specification get fixed.

Original:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - 2d8af1e0fbb120dd9101ca620c24c59e940d1be6
- DDS implementation:
  - Fast-RTPS
  - rclpy

#### Steps to reproduce issue

```python
from rosidl_parser.definition import NamespacedType
from rosidl_runtime_py.import_message import import_message_from_namespaced_type

message = import_message_from_namespaced_type(
            NamespacedType(['std_msgs', 'msg'], 'Bool'))
```

#### Expected behavior
Imports a `std_msgs/msg/Bool`
#### Actual behavior
Crash
```terminal
AttributeError: 'NamespacedType' object has no attribute 'value_type'
```
#### Additional information

Crash comes from [`import_message_from_namespaced_type`](https://github.com/ros2/rosidl_python/blob/master/rosidl_runtime_py/rosidl_runtime_py/import_message.py#L23) which expects a [`NamespacedType`](https://github.com/ros2/rosidl/blob/master/rosidl_parser/rosidl_parser/definition.py#L169) for input arg. `NamespacedType` expects as input args `Iterable[str], str`.

`NamespacedType` has base [`AbstractNestableType`](https://github.com/ros2/rosidl/blob/master/rosidl_parser/rosidl_parser/definition.py#L111) whereas the '`__slots__`' `value_type` comes from [`AbstractNestedType`](https://github.com/ros2/rosidl/blob/master/rosidl_parser/rosidl_parser/definition.py#L285). Am I missing something or is it a bug ?

mikaelarguedas:
Fixes #70 

Instead of changing the type specification, which would either require changing the function name (breaking API) or have the name and the behavior mismatch (current state), this modifies the function to operate on `NamespacedType` instead.
This will change behavior for consumers of the function.
If this is not desired, the less disruptive alternative is to change the function specification without changing the function name or the behavior.

Added a simple test to cover the uses of the function used in this package.

### DIFF
--- a/rosidl_runtime_py/rosidl_runtime_py/import_message.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/import_message.py
@@ -20,5 +20,5 @@ from rosidl_parser.definition import NamespacedType
 
 def import_message_from_namespaced_type(message_type: NamespacedType) -> Any:
     module = importlib.import_module(
-        '.'.join(message_type.value_type.namespaces))
-    return getattr(module, message_type.value_type.name)
+        '.'.join(message_type.namespaces))
+    return getattr(module, message_type.name)
--- a/rosidl_runtime_py/rosidl_runtime_py/import_message.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/import_message.py
@@ -14,11 +14,19 @@
 
 import importlib
 from typing import Any
+import warnings
 
 from rosidl_parser.definition import NamespacedType
 
 
 def import_message_from_namespaced_type(message_type: NamespacedType) -> Any:
+    if not isinstance(message_type, NamespacedType):
+        if hasattr(message_type, 'value_type'):
+            message_type = message_type.value_type
+            warnings.warn(
+                'Passing objects containing a NamespacedType is deprecated, '
+                'please pass a NamespacedType object directly',
+                DeprecationWarning)
     module = importlib.import_module(
         '.'.join(message_type.namespaces))
     return getattr(module, message_type.name)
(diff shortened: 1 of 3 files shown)


dirk-thomas:
For backward compatibility it would be nice to keep the previous usage pattern working.

We don't want to add a new function with a different name (the current name is just "right"). So how about the function checks if the passed `message_type` argument isn't of type `NamespacedType` and if that is the case and the variable has a `value_type` attribute then do `message_type = message_type.value_type`?

mikaelarguedas:
> So how about the function checks if the passed message_type argument isn't of type NamespacedType and if that is the case and the variable has a value_type attribute then do message_type = message_type.value_type?

I thought about this originally, the uncertainty I had was: "what would be the type of the argument in the signature?"

- Keep just `NamespacedType` even if the function can work with other types
- Use a specification matching more closely the accepted types: e.g. `Union[NamespacedType, AbstractNestedType]` in this case the change could simply be (though less permissive than accepting any type that has a attribute `value_type`):
```diff
-def import_message_from_namespaced_type(message_type: NamespacedType) -> Any:
+def import_message_from_namespaced_type(
+        message_type: Union[NamespacedType, AbstractNestedType]) -> Any:
+    if isinstance(message_type, AbstractNestedType):
+        message_type = message_type.value_type
```

- Something else?

dirk-thomas:
I would choose:

> Keep just `NamespacedType` even if the function can work with other types

I would consider the special case to keep current usage working as deprecated. Maybe add a deprecation in the case where accessing `value_type` so that existing usage can be updated and the logic removed in the future.

mikaelarguedas:
Sounds good, done in https://github.com/ros2/rosidl_python/pull/74/commits/000238809d14b317b2042c174699a6466c78befa

dirk-thomas:
Please rebase against the latest state on the target branch in order to run CI.

dirk-thomas:
[Build Status](https://ci.ros2.org/job/ci_linux/7679/)

PR #76, opened 2019-07-24
Remove non-package from ament_target_dependencies()

sloretz:
Also set _FOUND to trick `ament_target_dependencies()` call for test
interfaces

blocks ament/ament_cmake#180

### DIFF
--- a/rosidl_generator_py/CMakeLists.txt
+++ b/rosidl_generator_py/CMakeLists.txt
@@ -49,6 +49,7 @@ if(BUILD_TESTING)
   include(cmake/register_py.cmake)
   include(cmake/rosidl_generator_py_get_typesupports.cmake)
 
+  # Trick ament_target_dependencies() into thinking this package has been found
   set(rosidl_generator_py_FOUND "1")
   set(rosidl_generator_py_DIR "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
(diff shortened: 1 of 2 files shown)


rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -273,6 +273,8 @@ foreach(_typesupport_impl ${_typesupport_impls})
   ament_target_dependencies(${_target_name}
     "rosidl_generator_c"
     "rosidl_generator_py"
+  )
+  target_link_libraries(${_target_name}
```
sloretz: I'm not 100% sure this needs to go through `target_link_libraries()` since it works currently, and there's an `add_dependencies()` call above.

sloretz: I removed it since it wasn't needed before.

rosidl_generator_py/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ if(BUILD_TESTING)
   include(cmake/register_py.cmake)
   include(cmake/rosidl_generator_py_get_typesupports.cmake)
 
+  set(rosidl_generator_py_FOUND "1")
```
jacobperron: It's probably worth re-iterating the reason for this in a comment too

sloretz: Added comment in 8334b82

PR #78, opened 2019-08-05
Add utility functions to get interface from identifier

ivanpauno:
To avoid repeated code in some `ros2cli` packages.
See https://github.com/ros2/ros2cli/pull/301#discussion_r310283612.

### DIFF
--- a/rosidl_runtime_py/rosidl_runtime_py/utilities.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/utilities.py
@@ -19,26 +19,36 @@ from rosidl_parser.definition import NamespacedType
 
 def get_namespaced_type(path: Text):
     """Create a `NamespacedType` object from the full name of an interface."""
-    return __get_namespaced_type(path)
-
-
-def get_action_namespaced_type(path: Text):
-    """Create a `NamespacedType` object from the full name of an action."""
-    return __get_namespaced_type(path, 'action')
+    return _get_namespaced_type(path)
 
 
 def get_message_namespaced_type(path: Text):
     """Create a `NamespacedType` object from the full name of a message."""
-    return __get_namespaced_type(path, 'msg')
+    return _get_namespaced_type(path, 'msg')
 
 
 def get_service_namespaced_type(path: Text):
     """Create a `NamespacedType` object from the full name of a service."""
-    return __get_namespaced_type(path, 'srv')
+    return _get_namespaced_type(path, 'srv')
+
+
+def get_action_namespaced_type(path: Text):
+    """Create a `NamespacedType` object from the full name of an action."""
+    return _get_namespaced_type(path, 'action')
+
+
+def is_message(interface):
+    """Return `True` if `interface` is a message."""
+    return hasattr(interface, 'SLOT_TYPES')
+
+
+def is_service(interface):
+    """Return `True` if `interface` is a service."""
+    return hasattr(interface, 'Request') and hasattr(interface, 'Response')
 
 
 def is_action(interface):
-    """Return `True` if `action` is an action."""
+    """Return `True` if `interface` is an action."""
     return (
         hasattr(interface, 'Goal') and
         hasattr(interface, 'Result') and
@@ -46,17 +56,7 @@ def is_action(interface):
     )
 
 
-def is_service(interface):
-    """Return `True` if `srv` is a service."""
-    return hasattr(interface, 'Request') and hasattr(interface, 'Response')
-
-
-def is_message(interface):
-    """Return `True` if `msg` is a message."""
-    return hasattr(interface, 'SLOT_TYPES')
-
-
-def __get_namespaced_type(path: Text, default_namespace=None):
+def _get_namespaced_type(path: Text, default_namespace=None):
     namespace = path.split(sep='/')
     name = namespace[-1]
     namespace = namespace[0:-1]
--- a/rosidl_runtime_py/rosidl_runtime_py/utilities.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/utilities.py
@@ -15,6 +15,36 @@
 from typing import Text
 
 from rosidl_parser.definition import NamespacedType
+from rosidl_runtime_py.import_message import import_message_from_namespaced_type
+
+
+def get_interface(path: Text):
+    """Get an interface from its full name."""
+    return import_message_from_namespaced_type(get_namespaced_type(path))
+
+
+def get_message(path: Text):
+    """Get a message from its full name."""
+    interface = import_message_from_namespaced_type(get_message_namespaced_type(path))
+    if not is_message(interface):
+        raise RuntimeError('Expected the full name of a message')
+    return interface
+
+
+def get_service(path: Text):
+    """Get a service from its full name."""
+    interface = import_message_from_namespaced_type(get_service_namespaced_type(path))
+    if not is_service(interface):
+        raise RuntimeError('Expected the full name of a service')
+    return interface
+
+
+def get_action(path: Text):
+    """Get a message from its full name."""
+    interface = import_message_from_namespaced_type(get_action_namespaced_type(path))
+    if not is_action(interface):
+        raise RuntimeError('Expected the full name of an action')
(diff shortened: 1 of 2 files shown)


rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,61 @@
...
+    namespace = path.split(sep='/')
+    name = namespace[-1]
+    namespace = namespace[0:-1]
+    if len(namespace) == 1 and default_namespace is not None:
+        namespace.append(default_namespace)
+    return NamespacedType(namespace, name)
+
+
+def full_type_to_namespaced_type(path: Text):
+    """Create a `NamespacedType` object from the full name of an action."""
```
artivis:
```suggestion
    """Create a `NamespacedType` object from the full name of an interface."""
```

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def is_srv(srv):
+    """Return `True` if `srv` is a service."""
+    return hasattr(srv, 'Request')
+
+
+def is_msg(msg):
+    """Return `True` if `msg` is a message."""
+    return hasattr(msg, 'get_fields_and_field_types')
```
dirk-thomas: This is the old API from a message. Instead this should probably check for `SLOT_TYPES`.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,61 @@
...
+def __full_type_to_namespaced_type(path: Text, default_namespace=None):
+    namespace = path.split(sep='/')
+    name = namespace[-1]
+    namespace = namespace[0:-1]
+    if len(namespace) == 1 and default_namespace is not None:
+        namespace.append(default_namespace)
+    return NamespacedType(namespace, name)
+
+
+def full_type_to_namespaced_type(path: Text):
```
dirk-thomas: Why should all the functions start with `full_`? Maybe renamed this function to `get_namespaced_type` and the following ones `get_message|service|action_namespaced_type`.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def full_srv_type_to_namespaced_type(path: Text):
+    """Create a `NamespacedType` object from the full name of a service."""
+    return __full_type_to_namespaced_type(path, 'srv')
+
+
+def is_action(action):
+    """Return `True` if `action` is an action."""
+    return hasattr(action, 'Goal')
```
dirk-thomas: This should check all the attributes which are present in an action. I would also name the argument `interface` since it is unclear if it actually is an action.

Same for services below.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def is_srv(srv):
```
dirk-thomas: No abbreviation: `is_service`.

Same below for `is_msg`.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,61 @@
...
+from rosidl_parser.definition import NamespacedType
+
+
+def __full_type_to_namespaced_type(path: Text, default_namespace=None):
+    namespace = path.split(sep='/')
+    name = namespace[-1]
+    namespace = namespace[0:-1]
+    if len(namespace) == 1 and default_namespace is not None:
+        namespace.append(default_namespace)
+    return NamespacedType(namespace, name)
```
dirk-thomas: Please move this internal helper below all the functions using it.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,65 @@
...
+    return __get_namespaced_type(path, 'msg')
+
+
+def get_service_namespaced_type(path: Text):
+    """Create a `NamespacedType` object from the full name of a service."""
+    return __get_namespaced_type(path, 'srv')
+
+
+def is_action(interface):
+    """Return `True` if `action` is an action."""
```
dirk-thomas: Docblock should mention `interface` instead.

Same below.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def get_message_namespaced_type(path: Text):
+    """Create a `NamespacedType` object from the full name of a message."""
+    return __get_namespaced_type(path, 'msg')
+
+
+def get_service_namespaced_type(path: Text):
+    """Create a `NamespacedType` object from the full name of a service."""
+    return __get_namespaced_type(path, 'srv')
```
dirk-thomas: Nitpick: I would use the following order bases on increasingly higher concepts:
* `get_message_namespaced_type`
* `get_service_namespaced_type`
* `get_action_namespaced_type`

Same for the `is_*` functions below.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,65 @@
...
+    """Return `True` if `srv` is a service."""
+    return hasattr(interface, 'Request') and hasattr(interface, 'Response')
+
+
+def is_message(interface):
+    """Return `True` if `msg` is a message."""
+    return hasattr(interface, 'SLOT_TYPES')
+
+
+def __get_namespaced_type(path: Text, default_namespace=None):
```
dirk-thomas: This should have a single underscore prefix. There is no special semantic for double underscore outside of a class context.

ivanpauno:
@dirk-thomas Addressed your comments in https://github.com/ros2/rosidl_python/pull/78/commits/0f68df663a9e3d789701996e8d59c17be02e35ba.
I also added: `get_interface`, `get_message`, `get_service`, `get_action` in https://github.com/ros2/rosidl_python/pull/78/commits/55b011dd857ab0f06a282f8532d82e1686d53dd0, which I realized they will be usefull.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,95 @@
...
+def get_interface(path: Text):
+    """Get an interface from its full name."""
+    return import_message_from_namespaced_type(get_namespaced_type(path))
+
+
+def get_message(path: Text):
+    """Get a message from its full name."""
+    interface = import_message_from_namespaced_type(get_message_namespaced_type(path))
+    if not is_message(interface):
+        raise RuntimeError('Expected the full name of a message')
```
dirk-thomas: Maybe raise `ValueError` instead?

Also it would be good if the error message contains the argument within quotes.

rosidl_runtime_py/rosidl_runtime_py/utilities.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def get_namespaced_type(path: Text):
+    """Create a `NamespacedType` object from the full name of an interface."""
+    return _get_namespaced_type(path)
+
+
+def get_message_namespaced_type(path: Text):
+    """Create a `NamespacedType` object from the full name of a message."""
+    return _get_namespaced_type(path, 'msg')
```
dirk-thomas: Keyword arguments should be passed with the explicitly specified keyword: `default_namespace='msg'`.

ivanpauno:
CI (up to rosidl_generator
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7807/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3885/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6373/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7641/)

Issue #69, opened 2019-07-16
set_message_fields crash on numpy.ndarray


artivis:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - 2d8af1e0fbb120dd9101ca620c24c59e940d1be6
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

(A)
```terminal
ros2 topic pub /chatter geometry_msgs/msg/PoseWithCovarianceStamped \
> "header:
>   frame_id: ''
>   stamp:
>     nanosec: 0
>     sec: 0
> pose:
>   covariance: {}
>   pose:
>     orientation:
>       w: 0.0
>       x: 0.0
>       y: 0.0
>       z: 0.0
>     position:
>       x: 1.0
>       y: 1.0
>       z: 1.0"
```

(B)
```terminal
ros2 topic pub /chatter geometry_msgs/msg/PoseWithCovarianceStamped \
> "header:
>   frame_id: ''
>   stamp:
>     nanosec: 0
>     sec: 0
> pose:
>   covariance: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
>   pose:
>     orientation:
>       w: 0.0
>       x: 0.0
>       y: 0.0
>       z: 0.0
>     position:
>       x: 1.0
>       y: 1.0
>       z: 1.0"
```

#### Expected behavior

Publish a `PoseWithCovarianceStamped` message with the provided data.

#### Actual behavior
Crash
(A) 
```terminal
Failed to populate field: Required argument 'shape' (pos 1) not found
```

(B) 
```terminal
Failed to populate field: sequence too large; cannot be greater than 32
```

#### Additional information
[...]

PR #83, opened 2019-09-04
fix bug where numpy arrays were not understood in set_message.py

jcmonteiro:
This should solve #69

### DIFF
--- a/rosidl_runtime_py/rosidl_runtime_py/set_message.py
+++ b/rosidl_runtime_py/rosidl_runtime_py/set_message.py
@@ -13,14 +13,17 @@
 # limitations under the License.
 
 import array
+import numpy
 
 from typing import Any
 from typing import Dict
 
 from rosidl_parser.definition import AbstractNestedType
 from rosidl_parser.definition import NamespacedType
+from rosidl_parser.definition import BasicType
 from rosidl_runtime_py.convert import get_message_slot_types
 from rosidl_runtime_py.import_message import import_message_from_namespaced_type
+from rosidl_generator_py.generate_py_impl import SPECIAL_NESTED_BASIC_TYPES
 
 
 def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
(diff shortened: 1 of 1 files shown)


sloretz:
Thanks for the PR! Would you mind targeting it at the `master` branch instead? The workflow is fixes go to `master` first, and then are backported to `dashing` if possible.

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -38,6 +41,8 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
         field_type = type(field)
         if field_type is array.array:
             value = field_type(field.typecode, field_value)
+        elif field_type is numpy.ndarray:
+            value = field_type(shape=field.shape, dtype=field.dtype)
```
dirk-thomas: This part has already been fixed in #81.

jcmonteiro:
Missed that one. I believe we should close the PR then, right?

dirk-thomas:
It depends on what the other changes in this PR are for?

rosidl_runtime_py/rosidl_runtime_py/set_message.py
```diff
@@ -53,4 +58,9 @@ def set_message_fields(msg: Any, values: Dict[str, str]) -> None:
                     submsg = field_elem_type()
                     set_message_fields(submsg, value[n])
                     value[n] = submsg
+            if isinstance(rosidl_type.value_type, BasicType):
+                field_elem_type = rosidl_type.value_type.typename
+                field_elem_type = eval(SPECIAL_NESTED_BASIC_TYPES[field_elem_type]['dtype'])
+                for n in range(len(field_value)):
+                    value[n] = field_elem_type(field_value[n])
```
jcmonteiro: This last part I added because I thought it could help in the future with some custom or new message array type.

dirk-thomas: If you want to still add this part please update the PR to target `master` as well as describe how this part is being used with reproducible steps.

jcmonteiro: Thanks for the time. I'll close the PR since I believe this extra piece of code won't be of any use now.

Issue #95, opened 2019-12-03
Message field name same as python reserved keyword causes syntax error when importing generated message


ErikOrjehag:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - binaries
- Version or commit hash:
  - dashing
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

msg-file with the following contents
```
geometry_msgs/Point from
geometry_msgs/Point to
```

#### Expected behavior

Import message containing field named "from".

#### Actual behavior

Gets python runtime error:
```
  File "/opt/dredgebot_ws/install/dredgebot_behaviours/lib/python3.6/site-packages/dredgebot_behaviours/actions/make_coverage_plan.py", line 22, in <module>
    from dredgebot_msgs.msg import LineSegment
  File "/opt/dredgebot_ws/install/dredgebot_msgs/lib/python3.6/site-packages/dredgebot_msgs/msg/__init__.py", line 1, in <module>
    from dredgebot_msgs.msg._line_segment import LineSegment  # noqa: F401
  File "/opt/dredgebot_ws/install/dredgebot_msgs/lib/python3.6/site-packages/dredgebot_msgs/msg/_line_segment.py", line 82
    self.from = kwargs.get('from', Point())
            ^
SyntaxError: invalid syntax
```

#### Additional information

The message file is built using colcon without issues but when the message is imported in python we get an runtime error as described above. We suspect that the keyword "from" is a reserved symbol in python which results in a "SyntaxError". Changing the name "from"--->"start" in the msg-file fixes the problem. We suspect that this is a problem with all reserved keywords in python such as "for", "in", "is" etc but have not tried.

dirk-thomas:
Until a language specific name mangling is implemented to avoid keywords as described in ros2/design#172 your only option is to avoid using keywords in message definitions.

samiamlabs:
A warning when building/generating the messages until that is implemented would be nice though. I don't know how common it is to use reserved keywords as filelds but it can be a bit tricky to understad the cause of this bug based on the runtime error messages.

dirk-thomas:
> A warning when building/generating the messages until that is implemented would be nice though.

That would indeed be nice (but probably also require most of the complexity an actual solution will require). Any pull requests towards this would be more than welcome.

PR #96, opened 2019-12-04
Add warnings for reserved Python keywords in interface members

samiamlabs:
This is a temporary fix related to issue #95. Until a language-specific name mangling is implemented to avoid keywords as described in ros2/design#172, a warning would be nice.

I can add stderr prints for services and actions also if this pull request seems like a good idea.

Example build output from colcon:
```
Starting >>> dredgebot_msgs
Starting >>> rosidl_generator_py
Finished <<< rosidl_generator_py [7.05s]                                                               
[Processing: dredgebot_msgs]                             
--- stderr: dredgebot_msgs                                  
Member name "from" in the message "LatLonLineSegment" is a reserved keyword in Python and is not supported at the moment. Please use a different name.
---
Finished <<< dredgebot_msgs [39.4s]

Summary: 2 packages finished [39.5s]
  1 package had stderr output: dredgebot_msgs
```

### DIFF
--- a/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
+++ b/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ast import literal_eval
+import keyword
 import os
 import pathlib
+import sys
+from ast import literal_eval
 
 from rosidl_cmake import convert_camel_case_to_lower_case_underscore
 from rosidl_cmake import expand_template
@@ -31,6 +33,7 @@ from rosidl_parser.definition import FLOATING_POINT_TYPES
 from rosidl_parser.definition import IdlContent
 from rosidl_parser.definition import IdlLocator
 from rosidl_parser.definition import INTEGER_TYPES
+from rosidl_parser.definition import Message
 from rosidl_parser.definition import NamespacedType
 from rosidl_parser.parser import parse_idl_file
 
@@ -72,6 +75,18 @@ def generate_py(generator_arguments_file, typesupport_impls):
         locator = IdlLocator(*idl_parts)
         idl_file = parse_idl_file(locator)
         idl_content.elements += idl_file.content.elements
+
+    # NOTE(sam): remove when a language specific name mangling is implemented
+    for message in idl_content.get_elements_of_type(Message):
+        for member in message.structure.members:
+            msg_name = message.structure.namespaced_type.name
+            if (keyword.iskeyword(member.name)):
+                print(('Member name "{}" in the message "{}" is a '
+                       'reserved keyword in Python and is not supported '
+                       'at the moment. Please use a different name.')
+                      .format(member.name, msg_name), file=sys.stderr)
+
+    # TODO(sam): add reseved keyword warnings for services and actions
 
     for subfolder in modules.keys():
         with open(os.path.join(args['output_dir'], subfolder, '__init__.py'), 'w') as f:
--- a/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
+++ b/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from ast import literal_eval
 import keyword
 import os
 import pathlib
 import sys
-from ast import literal_eval
 
 from rosidl_cmake import convert_camel_case_to_lower_case_underscore
 from rosidl_cmake import expand_template
--- a/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
+++ b/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
@@ -26,6 +26,7 @@ from rosidl_cmake import read_generator_arguments
 from rosidl_parser.definition import AbstractGenericString
 from rosidl_parser.definition import AbstractNestedType
 from rosidl_parser.definition import AbstractSequence
+from rosidl_parser.definition import Action
 from rosidl_parser.definition import Array
 from rosidl_parser.definition import BasicType
 from rosidl_parser.definition import CHARACTER_TYPES
@@ -35,6 +36,7 @@ from rosidl_parser.definition import IdlLocator
 from rosidl_parser.definition import INTEGER_TYPES
 from rosidl_parser.definition import Message
 from rosidl_parser.definition import NamespacedType
+from rosidl_parser.definition import Service
 from rosidl_parser.parser import parse_idl_file
 
 SPECIAL_NESTED_BASIC_TYPES = {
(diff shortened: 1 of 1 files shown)


wjwwood:
@samiamlabs is this still a WIP?

samiamlabs:
I have not added error-prints for services or actions yet. If you think this PR makes sense I can fix that before merging.

Alternatively, we could just merge this PR and I can make a new one for actions and services.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -73,6 +76,18 @@ def generate_py(generator_arguments_file, typesupport_impls):
...
 
+    # NOTE(sam): remove when a language specific name mangling is implemented
+    for message in idl_content.get_elements_of_type(Message):
+        for member in message.structure.members:
+            msg_name = message.structure.namespaced_type.name
+            if (keyword.iskeyword(member.name)):
+                print(('Member name "{}" in the message "{}" is a '
+                       'reserved keyword in Python and is not supported '
+                       'at the moment. Please use a different name.')
+                      .format(member.name, msg_name), file=sys.stderr)
```
wjwwood: Please only use single indention steps and always a multiple of four spaces. Also one of the parenthesis pairs are redundant. Something like this instead:

```python
if (keyword.iskeyword(member.name)):
  print(
    'Member name "{}" in the message "{}" is a '
    'reserved keyword in Python and is not supported '
    'at the moment. Please use a different name.'
    .format(member.name, msg_name),
    file=sys.stderr)
```

Or if you want to keep the parenthesis make them do something specific like:

```python
if (keyword.iskeyword(member.name)):
  print(
    (
      'Member name "{}" in the message "{}" is a '
      'reserved keyword in Python and is not supported '
      'at the moment. Please use a different name.'
    ).format(member.name, msg_name),
    file=sys.stderr)
```

I prefer the first case though. The important thing though is to avoid this:

```
                print(('Member name "{}" in the message "{}" is a '
                       'reserved keyword in Python and is not supported '
              # ^-----^ this is more than 4 spaces
```

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -73,6 +76,18 @@ def generate_py(generator_arguments_file, typesupport_impls):
...
+
+    # TODO(sam): add reseved keyword warnings for services and actions
```
wjwwood: It would be better to do this now, but if you want to do it across more than one pull request I guess that would be ok too.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
wjwwood: Also, please use single quotes `'` to quote things being printed, and double quotes `"` to avoid needing escape characters. So `'Member name "{}" in the message "{}" is a '` -> `"Member name '{}' in the message '{}' is a "`.

wjwwood:
I'd advise you to fix it up to check in actions and services too before removing the WIP, but that's up to you. I do think this pr makes sense in general. @dirk-thomas do you think so too? My one concern is that this will warn on messages that already exist and cannot easily be changed, but maybe we already fixed that. I don't remember.

wjwwood:
ping @dirk-thomas 

Also, @samiamlabs do you think you'll have time to follow up on this? No pressure but it would help us triaging things to know.

samiamlabs:
I will try to find some time to work on this at the start of next week :)

samiamlabs:
Finally got around to do some more work on this.
Here is an example build output:

```
[Processing: rosidl_python_test_interfaces]                              
--- stderr: rosidl_python_test_interfaces                                  
Member name 'from' in the message 'ReservedKeywords' is a reserved keyword in Python and is not supported at the moment. Please use a different name.
Member name 'yield' in the message 'ReservedKeywords' is a reserved keyword in Python and is not supported at the moment. Please use a different name.
Member name 'is' in the service request 'ReservedKeywordsRequest' is a reserved keyword in Python and is not supported at the moment. Please use a different name.
Member name 'lambda' in the service response 'ReservedKeywordsResponse' is a reserved keyword in Python and is not supported at the moment. Please use a different name.
Member name 'nonlocal' in the action feedback 'ReservedKeywordsFeedback' is a reserved keyword in Python and is not supported at the moment. Please use a different name.
Member name 'global' in the action goal 'ReservedKeywordsGoal' is a reserved keyword in Python and is not supported at the moment. Please use a different name.
Member name 'pass' in the action result 'ReservedKeywordsResult' is a reserved keyword in Python and is not supported at the moment. Please use a different name.
---
Finished <<< rosidl_python_test_interfaces [30.7s]

```

I also set up a VSCode "Dev Container" for the nightly ROS2 docker image in case anyone is interested: https://github.com/samiamlabs/ros2-nightly-vscode/tree/reserved-keyword-warnings

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ast import literal_eval
+import keyword
 import os
 import pathlib
+import sys
+from ast import literal_eval
```
wjwwood: I'm surprised this is at the end of the list, but as long as the linters are happy it's fine.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9131/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4853/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7455/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9054/)

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
wjwwood: As I expected the linter doesn't care for this:

https://ci.ros2.org/job/ci_linux/9131/testReport/junit/(root)/projectroot/flake8/

ivanpauno:
@wjwwood friendly ping. Is this ready? Does something have to be addressed?

wjwwood:
There are unresolved linter failures, see: https://github.com/ros2/rosidl_python/pull/96#discussion_r370856833

samiamlabs:
I pushed a fix for the linter issues a while back @wjwwood 
Would be great if you can run CI again to make sure the errors are gone :)

ivanpauno:
CI (only fastrtps):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9256/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4940/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7562/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9175/)

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -73,6 +78,47 @@ def generate_py(generator_arguments_file, typesupport_impls):
         idl_file = parse_idl_file(locator)
         idl_content.elements += idl_file.content.elements
 
+    # NOTE(sam): remove when a language specific name mangling is implemented
+
+    def print_warning_if_reserved_keyword(member_name, interface_type, interface_name):
+        if (keyword.iskeyword(member.name)):
+            print(
+                "Member name '{}' in the {} '{}' is a "
```
ivanpauno: is it really not supported? or you just have to use everywhere:
```python3
getattr(msg, 'name_that_is_also_a_keyword')
```
which is ugly, but supported.

samiamlabs:
We noticed that using "from" in a message caused a runtime syntax error for python when importing the message. 
Some of the reserved keywords can probably be accessed using `getattr` without issues like this.
See #95

ivanpauno:
> We noticed that using "from" in a message caused a runtime syntax error for python when importing the message.
> Some of the reserved keywords can probably be accessed using `getattr` without issues like this.
> See #95

IMO, that could be fixed if we use everywhere in the generated files `getattr` instead of dot syntax.
If the current generated files raise an error when importing, we should rather fail than log a warning (maybe it's a bad idea, because it can generate problems in existing messages). The other option is to fix the places that makes the imported message to fail (using `getattr`), and just log a warning indicating the usage of a python keyword.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
wjwwood: Are you suggesting we just say "is not recommended", rather than "not supported"?

ivanpauno: Based on the below [answer](https://github.com/ros2/rosidl_python/pull/96#issuecomment-582893866), I think that not supported is the correct wording.

----

If we want to support message fields named as keywords in `Python`, an alternative to do name mangling is to use `getattr` everywhere in the generated files. The user would have to use the same in their programs (in case the field name was a keyword). Of course, this alternative is not viable in other programming languages (e.g.: c++), but for `Python` I actually prefer that than mangling (though it's arguable).

ivanpauno:
@samiamlabs Can you rebase? Most of the test failures are resolved in master.

samiamlabs:
I pushed a rebase with upstream master just now. Hope that helps with CI :)

ivanpauno:
> I pushed a rebase with upstream master just now. Hope that helps with CI :)

Thanks! Another round of CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9291/)

ivanpauno:
CI failure seems unrelated, merging!

PR #102, opened 2020-04-09
remove dependency on rmw_implementation

dirk-thomas:
Needed for ros2/rmw_dds_common#12

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_get_typesupports.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_get_typesupports.cmake
@@ -13,12 +13,13 @@
 # limitations under the License.
 
 macro(rosidl_generator_py_get_typesupports TYPESUPPORT_IMPLS)
-  set(TYPESUPPORT_IMPLS "")
-  ament_index_get_resources(_typesupport_impls "rosidl_typesupport_c")
-  foreach(_typesupport ${_typesupport_impls})
+  set(${TYPESUPPORT_IMPLS} "")
+  ament_index_get_resources(${TYPESUPPORT_IMPLS} "rosidl_typesupport_c")
+  list(APPEND ${TYPESUPPORT_IMPLS} "rosidl_typesupport_c")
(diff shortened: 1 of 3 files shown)


rosidl_generator_py/cmake/rosidl_generator_py_get_typesupports.cmake
```diff
@@ -12,19 +12,14 @@
...
-    list_append_unique(_typesupport_impls ${_typesupport_impl})
-  endif()
-endmacro()
-
 macro(rosidl_generator_py_get_typesupports TYPESUPPORT_IMPLS)
-  set(TYPESUPPORT_IMPLS "")
-  set(_typesupport_impls "")
-  call_for_each_rmw_implementation(accumulate_typesupports)
-  foreach(_typesupport_impl ${_typesupport_impls})
-    list_append_unique(TYPESUPPORT_IMPLS ${_typesupport_impl})
```
dirk-thomas, line 28: The previous logic was broken. The variable which name is passed in `TYPESUPPORT_IMPLS` is never set by the macro. This only happened to work if the caller expected the result in `_typesupport_impls`. :exploding_head:

ivanpauno: Oh, `cmake` macros are really tricky.

rosidl_generator_py/package.xml
```diff
@@ -36,8 +35,6 @@
   <test_depend>python3-pytest</test_depend>
   <test_depend>python_cmake_module</test_depend>
   <test_depend>rmw</test_depend>
-  <test_depend>rmw_implementation</test_depend>
```
jacobperron, line 39: I believe this change led to a failure on the [binary job](http://build.ros2.org/job/Fbin_uF64__rosidl_generator_py__ubuntu_focal_amd64__binary/). We're generating messages for tests, but there's no C typesupport available (previously transitively brought in by `rmw_implementation`).

Here is a proposed fix: https://github.com/ros2/rosidl_python/pull/110

PR #110, opened 2020-04-30
Add test dependency on rosidl_typesupport_c_packages

jacobperron:
Since we are generating interfaces for testing.

This is aiming to fix build failures on build.ros2.org, for example: http://build.ros2.org/job/Fbin_uF64__rosidl_generator_py__ubuntu_focal_amd64__binary/7

### DIFF
--- a/rosidl_generator_py/package.xml
+++ b/rosidl_generator_py/package.xml
@@ -36,7 +36,14 @@
   <test_depend>python_cmake_module</test_depend>
   <test_depend>rmw</test_depend>
   <test_depend>rosidl_cmake</test_depend>
-  <test_depend>rosidl_default_generators</test_depend>
+
+  <!--
+  Bloom does not support group_depend so entries below duplicate the group rosidl_typesupport_c_packages.
+  -->
+  <test_depend>rosidl_typesupport_connext_c</test_depend>
+  <test_depend>rosidl_typesupport_introspection_c</test_depend>
+  <test_depend>rosidl_typesupport_fastrtps_c</test_depend>
+  <!-- end of group dependencies added for bloom -->
   <test_depend>rosidl_generator_c</test_depend>
   <test_depend>rosidl_generator_cpp</test_depend>
   <test_depend>rosidl_parser</test_depend>
--- a/rosidl_generator_py/package.xml
+++ b/rosidl_generator_py/package.xml
@@ -36,6 +36,7 @@
   <test_depend>python_cmake_module</test_depend>
   <test_depend>rmw</test_depend>
   <test_depend>rosidl_cmake</test_depend>
+  <test_depend>rosidl_default_generators</test_depend>
   <test_depend>rosidl_generator_c</test_depend>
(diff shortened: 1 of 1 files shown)


rosidl_generator_py/package.xml
```diff
@@ -36,6 +36,14 @@
...
   <test_depend>rmw</test_depend>
   <test_depend>rosidl_cmake</test_depend>
+
+  <!--
+  Bloom does not support group_depend so entries below duplicate the group rosidl_typesupport_c_packages.
+  -->
+  <test_depend>rosidl_typesupport_connext_c</test_depend>
+  <test_depend>rosidl_typesupport_introspection_c</test_depend>
+  <test_depend>rosidl_typesupport_fastrtps_c</test_depend>
+  <!-- end of group dependencies added for bloom -->
```
jacobperron, line 46: Q: should we also add  `<group_depend>rosidl_typesupport_c_packages</group_depend>` ?

dirk-thomas: What for? Above we already `buildtool_export` depend on `rosidl_typesupport_c` which has such a group dependency.

jacobperron: Because it seems like it is necessary for this package to build (specifically the tests). It seems more correct than relying on it as a transitive dependency.

dirk-thomas: But a group dependency is not limited to tests but acts as a build as well as run dependency.

jacobperron: Yeah, so maybe it's best to leave it at the separate test_depend tags.

dirk-thomas: Should instead https://github.com/ros2/rosidl_typesupport/blob/3c15da21132af47d73a3fea00f2d99352e647f11/rosidl_typesupport_c/package.xml#L19-L20 be modified to include `rosidl_typesupport_fastrtps_c` and this package woulld only `test_depend` on `rosidl_typesupport_c` (as it also does `buildtool_export_depend` on)?

jacobperron: Maybe I'm wrong, but I don't think the issue is that `rosidl_typesupport_fastrtps_c` is missing. I think the issue is that no C typesupports are depended on here (I presume the build depends of `rosidl_typesupport_c` don't get installed when building `rosidl_generator_py`).

I guess adding one or more `<build_export_depend>` tags for C typesupport packages in `rosidl_typesupport_c` would fix the build though.

dirk-thomas: > I guess adding one or more `<build_export_depend>` tags for C typesupport packages in `rosidl_typesupport_c` would fix the build though.

That doesn't seem right from the perspective of that package.

How about we keep the patch as it is right now and see if it fixes the build as desired?

jacobperron: Sure. Considering that the PR job passes (when past PR jobs failed) I'm fairly confident it will work :crossed_fingers:

PR #116, opened 2020-06-23
Add pytest.ini so tests succeed locally.

clalancette:

This is part of ros2/ros2#951, which has more explanation and CI.


dirk-thomas:
Same comment as https://github.com/ros2/rosidl/pull/502#issuecomment-649652428

clalancette:
Fixed in 05542c6

Issue #124, opened 2021-02-10
Erroneous, unmatched closing parenthesis for char array members


spiderkeys:
- Operating System:
  - x64 Linux - Ubuntu 20.04
- Installation type:
  - From source
- Version or commit hash:
  - eb8d40621de46b5c9c242f3c56e59b97f81cc2d2
 
#### Steps to reproduce issue
1. Create a package that can generate types from IDL.
2. Create an IDL message file with a char[4] type as one of the struct fields.
3. Build the package and source the setup.bash
4. Use `ros2 topic pub <topic name> <any message type from package with IDL containing the char array> {}` or `ros2 topic echo <topic with type from that same package>` 
5. ros2cli tool will fail with a traceback

#### Description
Ultimately, the root cause (as indicated at the end of the error message below) is an unmatched closing parenthesis which is generated for the char[] type in _msg.py.em.

This does not:
 - Happen for other types that I have tested like `long[4]` and `short[4]`, only for `char[4]`
 - Happen for non-array `char`
 - Change based on array size (4 is just an example)

This issue seems to have been in the codebase for quite some time, but only started causing issues in a recent Rolling build that I performed about a week ago, so it seems like something changed that caused this to start happening at runtime. This may be related to this recent commit:
https://github.com/ros2/rosidl_python/commit/91df295f92547b3255b2ef948d4c36ebec6a1e2c#diff-0da744c3127964a8907eb35fc0b5c04e32eccd515d8cf7fb839ec80ebb82fa59

One clue is that the error for the offending struct happens even when publishing or subscribing to a topic with a different type than the char[] containing one within the same package, which I believe is caused by all types getting imported in `<some package>/msg/__init__.py`

#### Fix:
https://github.com/ros2/rosidl_python/pull/125

#### Example

myorg_msgs/msg/TestMessage.idl
```
module myorg_msgs {
  module msg {
    struct TestMessage {
       char [4];
    };
  };
};
```
Command/error:
```
spiderkeys@spiderdesk:/opt/mr/myorg$ ros2 topic pub hello myorg_msgs/msg/TestMessage {}
Traceback (most recent call last):
  File "/opt/ros/install/bin/ros2", line 33, in <module>
    sys.exit(load_entry_point('ros2cli==0.10.1', 'console_scripts', 'ros2')())
  File "/opt/ros/install/lib/python3.8/site-packages/ros2cli/cli.py", line 67, in main
    rc = extension.main(parser=parser, args=args)
[...]

PR #125, opened 2021-02-10
Removes erroneous unmatched closing parenthesis

spiderkeys:
Addresses: https://github.com/ros2/rosidl_python/issues/124


clalancette (changes requested):
This fix is clearly correct, so thanks for that.

It's clear that we aren't testing this functionality at all.  Can you please add a test for this?  I'll suggest adding a new message in the `msg` subdirectory here, and then adding in a new test into `test/test_interfaces.py`.

spiderkeys:
Will do 👍

spiderkeys:
@clalancette I've added some types (both .idl and .msg) that include a comprehensive set of built-in types, arrays, and sequences, as well as some basic test cases.

In doing so, I came across one additional error for char array types. See: https://github.com/ros2/rosidl_python/pull/125/commits/e925bb39eb69cb4b9dca9e1e65d46282a7e8d9d6

Let me know if this is what you had in mind. Happy to make any changes.

clalancette (commented):
This was exactly the kind of thing I had in mind, so thanks for that.

However, I'm now actually a bit confused on what is going on here.  Take your new `BuiltinTypesMsg.msg` message.  That tests all of the basic types and the array versions of the basic types.  But shouldn't that *already* have been done via the `BasicTypes` and `Array` message in `test_interfaces.py`?  That is, it seems like a lot of this is duplicating tests that are already there.  But clearly that isn't the case since this bug exists.  Do you have any thoughts on what is going on here?

sloretz:
>  That is, it seems like a lot of this is duplicating tests that are already there. But clearly that isn't the case since this bug exists. Do you have any thoughts on what is going on here?

@clalancette I think the difference is this adds a test for `char[2] char_array` to a message, which isn't currently covererd by `test_interface_files`. The rest of the fields added in tests in this PR may or may not be duplicating existing tests.

https://github.com/ros2/rosidl_python/pull/125/files#diff-4c54c46da9bed05c8c28519d8a90efacd5ac1bb2590c04b13aba8d893cfd7237R18-R33

spiderkeys:
Sorry, I kind of forgot about this PR.

As both @sloretz and @clalancette mention, there are likely a number of duplicate test cases - when I originally put this together, my thought process was to very explicitly capture most if not all basic permutations of the builtin types defined by the IDL spec. The proper way to remove the duplication is likely to move my Builtin* type files or a subset of their contents over to the `test_interface_files` repo so they all live in one place.

https://github.com/ros2/test_interface_files

I don't have much time in the next week or two to tackle this, but if you agree with the above change and someone could make the appropriate changes in `test_interface_files`, I could cut that aspect out of this PR so the fix can make its way in.

clalancette:
> @clalancette I think the difference is this adds a test for `char[2] char_array` to a message, which isn't currently covererd by `test_interface_files`. The rest of the fields added in tests in this PR may or may not be duplicating existing tests.

@sloretz Ah, thanks so much for pointing that out.  That was exactly what I was missing.

> As both @sloretz and @clalancette mention, there are likely a number of duplicate test cases - when I originally put this together, my thought process was to very explicitly capture most if not all basic permutations of the builtin types defined by the IDL spec.

@spiderkeys Yeah, understood.  I think I got bogged down in adding so many test cases, and they seemed like duplicated.  What I've done here now is to rebase this onto the latest, and then removed all of the test cases except for the one failure in b5ce601.  With that, I think this is easier to review and understand.  I'm happy with this now, so I'm going to go ahead and approve, but I'd appreciate a look from you just to make sure you agree with the direction here.

clalancette (approved):

clalancette:
@spiderkeys Oh, also if you could go in here and rebase to add `Signed-off-by` tags to every commit, that will make the DCO bot happy.  Thanks.

spiderkeys:
@clalancette I think I've just performed the rebase correctly to amend with signoffs, but let me know if something doesn't look right.

clalancette:
> @clalancette I think I've just performed the rebase correctly to amend with signoffs, but let me know if something doesn't look right.

The signoffs look good, thanks.

I ended up rebasing it again to have it based off of master.  With that, this looks good to me so I'll run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16127/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10756/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16458/)

clalancette:
All of the failures on Windows are known (and most of them should actually be fixed by now).  So I'm going to go ahead and merge this; thanks again for the contribution!

flynneva:
@clalancette following up on this topic - has this bug-fix been backported to other ros2 distros as well? Would you be willing to take PR's to backport it? Would be great to get this fixed in all distros rather than just `rolling`...especially with the transition to jammy coming up soon I think.

clalancette:
> @clalancette following up on this topic - has this bug-fix been backported to other ros2 distros as well? Would you be willing to take PR's to backport it? Would be great to get this fixed in all distros rather than just `rolling`...especially with the transition to jammy coming up soon I think.

The transition to Jammy (and Humble) is done now.  In terms of this being backported, no, we haven't done it yet.  I'll kick it off to see if the bot can do it.

clalancette:
@Mergifyio backport foxy galactic

PR #129, opened 2021-04-28
Added optimization for copying arrays of simple types from python to C using buffer protocol

ksuszka:
This is rather naive approach to improve perfomance of copying large amount of data from python to C.
I didn't remove old code, but instead I've added a check if data can be accessed using buffer protocol, and if the answer is yes then a sequence is copied in single step, if the answer is no, then fallback to previous method is used.

It seems to work for ROS2 version of carla_ros_bridge where we discovered this issues. It seems to be the same issue as described in https://github.com/ros2/rclpy/issues/763

PR #130, opened 2021-05-22
Fix int32[] and uint32[] 32 bit sequence types only accepting 16 bit sequences

sloretz:
I noticed this bug while reviewing #129. There is a bug with the types of `int32[]` and `uint32[]` message fields where 32 bit sequences are being stored in types that only accepts 16bit values. The tests look like they're covering this case, but actually they aren't because the comparison of `array.array()` instances is not intuitive.

`array.array()` comparisons don't consider typecode

```python3
>>> array.array('B') == array.array('f')
True
```

`array.array()` comparisons seem to check equality of the sequences they were created with, not the bytes they output

```python3
>>> aI =  array.array('I', [0, 1, 4294967295])
>>> aL = array.array('L', [0, 1, 4294967295])
>>> aI == aL
True
>>> aI.tobytes() == aL.tobytes()
False
```

This should be backported to Foxy and Galactic.

### DIFF
--- a/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
+++ b/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
@@ -46,8 +46,8 @@ SPECIAL_NESTED_BASIC_TYPES = {
     'uint8': {'dtype': 'numpy.uint8', 'type_code': 'B'},
     'int16': {'dtype': 'numpy.int16', 'type_code': 'h'},
     'uint16': {'dtype': 'numpy.uint16', 'type_code': 'H'},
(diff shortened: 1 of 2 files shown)


sloretz:
Full CI to see if fixing the types breaks anything downstream

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14678/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9430/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12342/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14821/)

clalancette:
This is a good catch, but unfortunately CI seems pretty cranky with this change :(.

PR #131, opened 2021-05-26
Install generated Python interfaces in a Python package

hidmic:
This makes Python interfaces discoverable through `pkg_resources` and `importlib.metadata`. 

Full CI to be safe:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14692/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9444/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12357/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14835/)

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -121,19 +121,7 @@ rosidl_write_generator_arguments(
 )
 
 if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
-  ament_python_install_module("${_output_path}/__init__.py"
-    DESTINATION_SUFFIX "${PROJECT_NAME}"
-  )
-
-  # TODO(esteve): replace this with ament_python_install_module and allow a list
-  # of modules to be passed instead of iterating over _generated_py_files
-  # See https://github.com/ros2/rosidl/issues/89
-  foreach(_generated_py_dir ${_generated_py_dirs})
-    install(DIRECTORY "${_output_path}/${_generated_py_dir}/"
-      DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}/${_generated_py_dir}"
-      PATTERN "*.py"
-    )
-  endforeach()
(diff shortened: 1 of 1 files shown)


sloretz (approved):

jacobperron:
This change prevents a package from building if it both generates interfaces and installs a Python package. For example, I have a package that is not longer building since it does something like this:

```cmake
rosidl_generate_interfaces(${PROJECT_NAME}_msgs
  ...
)

ament_python_install_package(${PROJECT_NAME})
```

I see an error like:

```
CMake Error at /home/jacob/ros2-linux/share/ament_cmake_python/cmake/ament_python_install_package.cmake:108 (add_custom_target):
  add_custom_target cannot create target
  "ament_cmake_python_symlink_my_package" because another target with the
  same name already exists.  The existing target is a custom target created
  in source directory
  ...
```

hidmic:
I can't say I didn't see this one coming. A user-defined `ament_python_install_package(${PROJECT_NAME})` contends with `rosidl_generate_interfaces()` to install the `${PROJECT_NAME}` Python package. That was not a problem before because there was no Python package (with metadata, discoverable). Python sources were installed like plain files, incrementally.

Maybe there is a way for an `ament_python_install_package(${PROJECT_NAME})` call to extend or augment previous ones, somehow. We would not be able to tell if that was on purpose or not though. 
WDYT @ros2/team ?

jacobperron:
> Maybe there is a way for an `ament_python_install_package(${PROJECT_NAME})` call to extend or augment previous ones, somehow. We would not be able to tell if that was on purpose or not though.

If it's possible, we could make it an option to extend a previous call to `ament_python_install_package`. This way we would know if someone is doing this on purpose. Then the error message could also be improved to suggest using the "extend" option if it detects multiple calls to `ament_python_install_package` with the same target name.

jtbandes:
I'm running into this error in rosbridge_suite as well, when trying to build for Rolling: https://github.com/RobotWebTools/rosbridge_suite/issues/581

Is there any recommendation for how to fix or work around this issue?

jtbandes:
Filed a new issue to track rather than discussing on an old merged PR: https://github.com/ros2/rosidl_python/issues/141

PR #144, opened 2021-09-13
remove extra parenthesis, closes #143

### LINKED ISSUE #143: extra parenthesis in generated file
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - galactic
- DDS implementation:
  - NA
  - N/A

#### Steps to reproduce issue

```
cd colcon_ws/src
git clone https://github.com/ros-realtime/reference-system.git

cd colcon_ws
colcon build
colcon test
```

the logs for `reference_interfaces` should show the error during the flake8 tests.

#### Expected behavior

generated files do not throw lint errors and can be run without error

#### Actual behavior

Generated file has this error when passed into the python interpreter:

```
build/reference_interfaces/rosidl_generator_py/reference_interfaces/msg$ python3 _transmission_stats.py 
  File "_transmission_stats.py", line 167
    all(val >= 0 and val) < 256 for val in value)), \
                                                 ^
SyntaxError: unmatched ')'
```

it looks like an extra parenthesis was added to the file.

#### Additional information

this is [the offending idl file](https://github.com/ros-realtime/reference-system/blob/main/reference_interfaces/msg/TransmissionStats.idl)

UPDATE: update/fix broken links

flynneva:
closes #143

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -495,7 +495,7 @@ bound = 2**nbits
                  all(val >= 0 and val < @(bound) for val in value)), \
 @{assert_msg_suffixes.append('and each unsigned integer in [0, %d]' % (bound - 1))}@
 @[    elif isinstance(type_, BasicType) and type_.typename == 'char']@
-                 all(val >= 0 and val) < 256 for val in value)), \
+                 all(ord(val) >= 0 and ord(val) < 256 for val in value)), \
 @{assert_msg_suffixes.append('and each char in [0, 255]')}@
 @[    elif isinstance(type_, BasicType) and type_.typename in FLOATING_POINT_TYPES]@
 @[      if type_.typename == "float"]@


sloretz (approved):
LGTM with green CI

sloretz:
CI (repos: `https://gist.githubusercontent.com/sloretz/5907cc90bb01ccb11c780949f11454a0/raw/75a1a69d7acb30dc6549d66c79801aed5d6cc09e/ros2.repos` build: `--packages-above-and-dependencies rosidl_generator_py` test: `--packages-above rosidl_generator_py`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15206/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9926/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12886/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15388/)

flynneva:
@sloretz tested locally and it seems like I deleted the wrong (?) parenthesis here....one sec

sloretz:
Interesting, I didn't test locally but visually it seems like it would delete the correct parenthesis to me

relevant part of generated code without this PR.
```
    def node_name(self, value):
        if __debug__:
            from collections.abc import Sequence
            assert \
                ((isinstance(value, Sequence) or
                  isinstance(value, Set) or
                  isinstance(value, UserList)) and
                 not isinstance(value, str) and
                 not isinstance(value, UserString) and
                 len(value) == 56 and
                 all(isinstance(v, str) for v in value) and
                 all(val >= 0 and val) < 256 for val in value)), \
                "The 'node_name' field must be a set or sequence with length 56 and each value of type 'str' and each char in [0, 255]"
        self._node_name = value
```

flynneva:
@sloretz yes same here, thought it made sense just looking at it....however when testing there is this error thrown with this change:

```
File "/opt/ros/galactic/lib/python3.8/site-packages/rclpy/executors.py", line 333, in _take_subscription
[1.902s] 1:     msg_info = _rclpy.rclpy_take(capsule, sub.msg_type, sub.raw)
[1.903s] 1: RuntimeError: make_tuple(): unable to convert arguments to Python object (compile in debug mode for details)
```

if I switch which parenthesis to delete, I am not seeing an error - so like this:

```
all(val >= 0 and val) < 256 for val in value), \
```

^^ note there is only one parenthesis at the end now....and this doesnt thrown an error and seems to be working ok I think.

flynneva:
@sloretz just testing the logic so im not going crazy, this should work I think:

```
def checkChar(value):
    return all((val >= 0 and val < 256) for val in value)


if __name__ == "__main__":
    name_true = [0, 1, 10, 5, 250]
    name_false = [0, 1, 300, 5, 250]
    print(checkChar(name_true))
    print(checkChar(name_false))
```

the above prints out:

```
True
False
```

Im going to push an update to this way of writing the `all` line. im guessing the error im seeing is a result of something else then.

flynneva:
@sloretz switching the problematic line to this seemed to fix all the problems I was seeing while also still doing a valuable test on each value in the char array:

```
all(len(val) == 1 for val in value))
```

flynneva:
> all(val >= 0 and val < 256 for val in value)

so this line compiles fine, however when testing it throws the tuple error:

```
RuntimeError: make_tuple(): unable to convert arguments to Python object (compile in debug mode for details)
```

im not sure if that means there is another problem down the path, or if that line is causing it.

clalancette:
It looks to me like that is a separate (though obviously related) problem.  At least, with this patch in place, the generated code looks more reasonable to me, so I think we need this patch as-is at a minimum.

However, I will point out that we don't appear to have any tests for this situation in the core ROS 2 code (which is why it has lingered for so long).  So I would suggest that we add a test for this situation as well.

flynneva:
@clalancette @sloretz let me know what direction you'd like to go or if you'd want to add tests to this PR (?) before merging.

clalancette:
> @clalancette @sloretz let me know what direction you'd like to go or if you'd want to add tests to this PR (?) before merging.

Yeah, let's add some tests here to start with.  That way we'll at least exercise this code.  If it turns out that the tests run into the same issue you are running into locally, then we'll have to dig deeper and come up with another fix as well.

sloretz:
Tests are great; they're a bit of work when testing the message generators. If you're willing to do it it would be a big benefit.

I would recommend to do it by creating a new message in `test_interface_files` with some descriptive name like `FixedSequences.msg`. See the sum of these PRs for how to add a new message, and all the places that need to be updated when `test_interface_files` gets a new message.

* https://github.com/ros2/test_interface_files/pull/14
* https://github.com/ros2/rmw_implementation/pull/193
* https://github.com/ros2/rcl_interfaces/pull/123
* https://github.com/ros2/rcl_interfaces/pull/125
* https://github.com/ros2/system_tests/pull/481

flynneva:
also just an fyi but [you can see our CI failing](https://github.com/ros-realtime/reference-system-autoware/runs/3602963839?check_suite_focus=true#step:5:510) due to the error this PR is trying to fix. will try and get a basic test for this implemented soon so we can get this merged in

flynneva:
@sloretz or @clalancette, any advice on how to get colcon to use a local `rosidl_generator_py` package rather than the default thats installed in `/opt/ros`?  for some reason including it in my `colcon_ws` isnt doing the trick...and the `test_msgs` package is still using the binary install.

flynneva:
@clalancette @sloretz figued it out why we were seeing this error. it DOES have to do with this line and nothing else.

in python to compare a `char` to a number you need to use the `ord()` function to return the `ASCII` value of that char.

so the line should be:

```
all(ord(val) >= 0 and ord(val) < 256 for val in value))
```

flynneva:
here is the final generated file for this message type, note the difference from the above

```
    def node_name(self, value):
        if __debug__:
            from collections.abc import Sequence
            assert \
                ((isinstance(value, Sequence) or
                  isinstance(value, Set) or
                  isinstance(value, UserList)) and
                 not isinstance(value, str) and
                 not isinstance(value, UserString) and
                 len(value) == 56 and
                 all(isinstance(v, str) for v in value) and
                 all(ord(val) >= 0 and ord(val) < 256 for val in value)), \
                "The 'node_name' field must be a set or sequence with length 56 and each value of type 'str' and each char in [0, 255]"
```

ideally I think we could add in one more check as well to make sure each `char` is just one byte....something like this would do:

```
all(len(val) == 1 for val in value) and
```

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/call-cmake-add-custom-command-after-rosidl-generator-py-is-done/22299/1

clalancette:
The changes look reasonable to me, though I would want to wait to merge this until we have the test ready.

flynneva:
@clalancette to use this modified version of the package locally, can I just clone it into my workspace and itll be used instead of the binary installed one with the error?

clalancette:
> Or do I also then need to rebuild all the packages that use it as well?

This package is generating python code, so to be totally safe you should rebuild everything above it.  On the other hand, since we don't seem to be using this particular code sequence in the core, you may be able to get away with only rebuilding the message packages that you have that need it.  I'm not sure.

flynneva:
@sloretz @clalancette just an fyi because I just realized it now.....down a [few lines in the same file](https://github.com/ros2/rosidl_python/blob/master/rosidl_generator_py/resource/_msg.py.em#L534) is the solution already implemented.

thought it was funny seeing it there, wish I would've noticed it more quickly! 😄

clalancette:
Also, FYI, #125 fixes this exact thing, we just never merged it because we never came to consensus on the testing aspect.

PR #160, opened 2022-05-02
Use modern cmake targets to avoid absolute paths to appear in binary archives

ivanpauno:
* Move logic to find Numpy headers to a module and export a target from there.

* Use the new NumpyHeaders::NumpyHeaders and PythonExtra::PythonExtra to avoid absolute paths to appear in windows binary archives.

This partially fixes https://github.com/ros2/rclcpp/issues/1688.
Depends on https://github.com/ros2/python_cmake_module/pull/11.


ivanpauno:
Windows packaging CI after this change: [Build Status](http://ci.ros2.org/job/ci_packaging_windows/198/)

ivanpauno:
Let's see if I have more luck now: [Build Status](http://ci.ros2.org/job/ci_packaging_windows/199/)

ivanpauno:
Third time is the charm (?): [Build Status](http://ci.ros2.org/job/ci_packaging_windows/201/)

ivanpauno:
I downloaded the binary archive, and it works!!!
Now there're a lot less absolute paths in the provided binaries.

rosidl_generator_py/cmake/FindNumpyHeaders.cmake
```diff
@@ -0,0 +1,62 @@
...
+)
+
+if(NumpyHeaders_FOUND)
+    set(NumpyHeaders_INCLUDE_DIRS ${NumpyHeaders_INCLUDE_DIR})
+
+    add_library(NumpyHeaders::NumpyHeaders INTERFACE IMPORTED)
+    set_target_properties(NumpyHeaders::NumpyHeaders PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${NumpyHeaders_INCLUDE_DIR}"
+    )
+endif()
```
clalancette, line 62: Nit: add trailing newline here.

clalancette (approved):
Overall, this looks good to me with green CI on all platforms (including RHEL and Windows Debug).  I'd like to get a second opinion from @sloretz before we merge, though.

That said, this seems pretty risky to me to put into Humble.  This is deep in the heart of the message generation, and we don't know the unintended side effects of this.  So my suggestion is that we do indeed put it into Rolling, but we do not backport to Humble.  I'd be happy to hear other opinions, though.

sloretz (changes requested):
In [CMake 3.14 and above, `Python3::NumPy`](https://cmake.org/cmake/help/latest/module/FindPython3.html) can be used, and looking at `webos OSE` I think we can use that here https://github.com/ros-infrastructure/rep/pull/351

ivanpauno:
> In [CMake 3.14 and above, Python3::NumPy](https://cmake.org/cmake/help/latest/module/FindPython3.html) can be used, and looking at webos OSE I think we can use that here https://github.com/ros-infrastructure/rep/pull/351

Thanks, I'm not sure why I didn't see that before.

ivanpauno:
> That said, this seems pretty risky to me to put into Humble. This is deep in the heart of the message generation, and we don't know the unintended side effects of this.

I think it's safe to backport.
The libraries generated here are only loaded from python to bind with the rosidl_generator_c interface, but nothing depends on them, so we actually don't even need to export the cmake targets.

ivanpauno:
Windows packaging job: [Build Status](http://ci.ros2.org/job/ci_packaging_windows/201/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16688/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11275/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17089/)

sloretz:
* Linux re-run [Build Status](https://ci.ros2.org/job/ci_linux/16689/)

sloretz (approved):
LGTM for Rolling and Humble backport when CI comes back green.

ivanpauno:
@mergifyio backport humble

PR #149, opened 2021-12-10
Fix rosidl_generator_py assuming incorect library names

sloretz:
Requires ament/ament_cmake#371

rosidl_generator_py was assuming all interface packages being depended
upon call their interface generation target name `"${PROJECT_NAME}"`, but
that's an incorrect assumption. The target name is a required argument
to `rosidl_generate_interfaces()`. This fixes it by exporting the python
library target using an existing CMake macro for that purpose called
`rosidl_export_typesupport_targets()` and `ament_export_targets()`.

I works by adding a variable
`"${PROJECT_NAME}_TARGETS__rosidl_generator_py"` which is set when the
interface package is `find_package()`d. That variable contains the
targets generated by `rosidl_generator_py` so that downstream interface
packages can depend on it.

Since targets are being exported, a bug where build time include paths were marked `PUBLIC` also needed to be fixed.

I discovered this bug while working on ament/ament_cmake#365, and I need it fixed for ros2/rcl_interfaces#133

PR #164, opened 2022-05-16
Don't export the generated libraries as targets anymore

ivanpauno:
Nothing should depend on them (they are meant only to be loaded as python extensions).

Issue was introduced in https://github.com/ros2/rosidl_python/pull/160.

This was causing build issues in rhel https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_release/1139/console#console-section-12.
It didn't cause issues in other platforms as they realize the library they're linking agains isn't being used for anything, so they don't look at the missing symbols.
But rhel linker seems to be more picky.

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -273,14 +273,9 @@ if(WIN32)
 endif()
 if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
   install(TARGETS ${_target_name_lib}
-    EXPORT export_${_target_name_lib}
     ARCHIVE DESTINATION lib
     LIBRARY DESTINATION lib
     RUNTIME DESTINATION bin)
-
-  # Export this target so downstream interface packages can depend on it
-  rosidl_export_typesupport_targets("${rosidl_generator_py_suffix}" "${_target_name_lib}")
-  ament_export_targets(export_${_target_name_lib})
 endif()
 
 if(BUILD_TESTING AND rosidl_generate_interfaces_ADD_LINTER_TESTS)


ivanpauno:
* rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/230/)

ivanpauno:
Using the NumPy component of FindPython3 introduced an issue in our Windows debug build, I will fix that one separately https://ci.ros2.org/view/nightly/job/nightly_win_deb/2357/console.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -273,14 +273,9 @@ if(WIN32)
 endif()
 if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
   install(TARGETS ${_target_name_lib}
-    EXPORT export_${_target_name_lib}
     ARCHIVE DESTINATION lib
     LIBRARY DESTINATION lib
     RUNTIME DESTINATION bin)
-
-  # Export this target so downstream interface packages can depend on it
-  rosidl_export_typesupport_targets("${rosidl_generator_py_suffix}" "${_target_name_lib}")
```
sloretz, line 282: Downstream interface packages depend on them. I'm not 100% sure they have to depend on them though.

https://github.com/ros2/rosidl_python/blob/3ce24baa074de225a314cbe8a286734e63482c96/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake#L264

That's using the special rosidl suffix'd targets variable. How about removing only the `ament_export_targets()`? That way downstream targets won't accidentally link against them if using `pkg_name_TARGETS` or `ament_target_dependencies()`.

clalancette:
Just as an FYI, before this PR asan builds are also broken (that is, `colcon build --mixin asan-gcc` fails to link).  With this PR, asan builds now work again.  I haven't tested @sloretz 's suggestion yet.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
ivanpauno: > Downstream interface packages depend on them. I'm not 100% sure they have to depend on them though.

Ah, I see.
I'm pretty sure that's not needed, but let me double check.

ivanpauno:
* rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/231/)

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
ivanpauno: > I checked the generated code, and yes a downstream interface package depends on the generated library.

Based on this, it seems that a bunch of stuff that I moved to `PRIVATE` in https://github.com/ros2/rosidl_python/pull/160 is not correct.
Why I did that is because `ament_export_dependencies()` doesn't allow you to specify components, so I worked that arround by not having to export anything at all.
But in retrospective that seems incorrect.

I have now exported the dependency of FindPython3 manually, creating a config file.

> That's using the special rosidl suffix'd targets variable. How about removing only the ament_export_targets()? That way downstream targets won't accidentally link against them if using pkg_name_TARGETS or ament_target_dependencies().

Unfortunately `rosidl_export_typesupport_targets()` requires you to also `ament_export_targets()`.
It would be nice if `ament_export_targets()` had an option to not append the target to pkg_name_TARGETS.
But considering that I have undone most of what I did in #160, I think we don't need to worry about that now.

ivanpauno:
* rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/232/)

wolfv:
This would be a very useful change for our cross-compiled builds for macOS ARM64 in RoboStack (currently facing an issue with absolute paths to numpy include dir that are exported from here): https://github.com/RoboStack/ros-humble/runs/7052219327?check_suite_focus=true

tobiasneumann:
What is the status of this PR?
As @wolfv said, this package is broken for cross compiling (I think the problem was introduced with 0f43ffc490b7e4ec08022361bced5fa274246dfd) and this might fix it?

clalancette:
> What is the status of this PR?

It's on hiatus at the moment.  At the very least it needs to be rebased and retested with the latest, and then it would also need to be reviewed.

> As @wolfv said, this package is broken for cross compiling (I think the problem was introduced with [0f43ffc](https://github.com/ros2/rosidl_python/commit/0f43ffc490b7e4ec08022361bced5fa274246dfd)) and this might fix it?

Can you please confirm that it actually does fix it?

tobiasneumann:
I'm working with `meta-ros` and there with the humble PR 4, hence I've tested it based on `0.14.4`.
To avoid conflicts I reapplied https://github.com/ros2/rosidl_python/pull/160 (after it was reverted in the master) and then this PR.

This does not work.

It fails with `cmake` for `rmw-cyclonedds-cpp` not been able to find python3-numpy correctly.
> Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found version "3.10.4")

For context (not directly relevant for this PR), `0.14.4` with https://github.com/ros2/rosidl_python/pull/149 reverted works.

PR #165, opened 2022-05-16
Fix how to find Numpy headers for Windows debug

ivanpauno:
Issue was introduced in https://github.com/ros2/rosidl_python/pull/160.
Example issue: https://ci.ros2.org/view/nightly/job/nightly_win_deb/2357/console.

I'm honestly guessing, worst case I will restore the logic to find Numpy headers using before #160.

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -21,6 +21,11 @@ find_package(PythonInterp 3.6 REQUIRED)
 
 find_package(python_cmake_module REQUIRED)
 find_package(PythonExtra MODULE REQUIRED)
+set(_PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE})
+if(WIN32 AND "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+  set(PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE_DEBUG})
+endif()
+
 find_package(Python3 REQUIRED COMPONENTS Development NumPy)
 
 # Get a list of typesupport implementations from valid rmw implementations.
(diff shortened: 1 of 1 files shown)


ivanpauno:
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/17106/)

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -21,6 +21,11 @@ find_package(PythonInterp 3.6 REQUIRED)
 
 find_package(python_cmake_module REQUIRED)
 find_package(PythonExtra MODULE REQUIRED)
+set(_PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE})
+if(WIN32 AND "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+  set(PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE_DEBUG})
```
sloretz, line 26: I'd guess setting `Python3_EXECUTABLE` to `PYTHON_EXECUTABLE_DEBUG` before `find_package(Python3 ...)` would allow using the the debug version of NumPY on Windows

https://cmake.org/cmake/help/latest/module/FindPython3.html#artifacts-specification

sloretz: Similar to here https://github.com/ros2/rclpy/blob/7ffe94865f38c67812799bf5751a8f0f3f918c25/rclpy/CMakeLists.txt#L17-L24

ivanpauno: That's exactly what I wanted to do, but I didn't do it correctly.

ivanpauno:
I will also fix this in https://github.com/ros2/rosidl_python/pull/164, as managin two PRs is making things more complicated

PR #187, opened 2022-10-26
use ament_python_install_package when package is not installed

### LINKED ISSUE #141: Build error when using ament_python_install_package() and generated interfaces
_Originally posted by @jacobperron in https://github.com/ros2/rosidl_python/issues/131#issuecomment-856221882_:

> This change [#131] prevents a package from building if it both generates interfaces and installs a Python package. For example, I have a package that is not longer building since it does something like this:
> 
> ```cmake
> rosidl_generate_interfaces(${PROJECT_NAME}_msgs
>   ...
> )
> 
> ament_python_install_package(${PROJECT_NAME})
> ```
> 
> I see an error like:
> 
> ```
> CMake Error at /home/jacob/ros2-linux/share/ament_cmake_python/cmake/ament_python_install_package.cmake:108 (add_custom_target):
>   add_custom_target cannot create target
>   "ament_cmake_python_symlink_my_package" because another target with the
>   same name already exists.  The existing target is a custom target created
>   in source directory
>   ...
> ```

I'm running into this issue in rosbridge_suite as well when building for Rolling: https://github.com/RobotWebTools/rosbridge_suite/issues/581

One thing I don't understand is why this error appears for [rosapi](https://github.com/RobotWebTools/rosbridge_suite/blob/bfe5cabc3ad6c77635eb500174fd57222b6aa19d/rosapi/CMakeLists.txt#L12-L43), but not [rosbridge_library](https://github.com/RobotWebTools/rosbridge_suite/blob/bfe5cabc3ad6c77635eb500174fd57222b6aa19d/rosbridge_library/CMakeLists.txt#L11-L36).

I'm a CMake noob, but would it be possible to simply pass a different, unique target name on this line instead of just `${PROJECT_NAME}`?

https://github.com/ros2/rosidl_python/blob/191a1435be9df09a1c2945e8558300c566e4219d/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake#L123-L125

(Meta-question: is there anything we could have done in either the rosbridge_suite repo or this repo to catch this kind of failure sooner, and not have to wait for another Rolling release to incorporate a fix into rosbridge_suite? Asking because I'm still learning about the release process, having recently taken on some maintenance of rosbridge_suite for ROS 2 compatibility.)

knorth55:
this PR fix #141 
with this PR, we can use `rosidl_generator_interfaces` and `ament_python_install_package` in the same CMakeLists.txt.

**But the order is important.**
as https://github.com/ros-drivers/audio_common/pull/212, we need to `ament_python_install_package` first and `rosidl_generate_interfaces` later.

```cmake
ament_python_install_package(${PROJECT_NAME})

rosidl_generate_interfaces(${PROJECT_NAME}
  ${msg_files}
  ${action_files}
  DEPENDENCIES
    action_msgs
    audio_common_msgs
    builtin_interfaces
)
```

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -122,10 +122,6 @@ rosidl_write_generator_arguments(
 
 if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
   if(${PROJECT_NAME} IN_LIST AMENT_CMAKE_PYTHON_INSTALL_INSTALLED_NAMES)
-    ament_python_install_module("${_output_path}/__init__.py"
-      DESTINATION_SUFFIX "${PROJECT_NAME}"
-    )
-
     # TODO(esteve): replace this with ament_python_install_module and allow a list
     # of modules to be passed instead of iterating over _generated_py_files
     # See https://github.com/ros2/rosidl/issues/89


knorth55:
@clalancette kindly ping. 
I made a PR to avoid the error reported in #141 .
Could you check this PR?

quarkytale:
This makes sense to me based on the conversation in the linked issue.
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17894/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12462/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18568/)

knorth55:
I update this PR to follow this comment.
thank you @MarcoMagriDev 
https://github.com/ros2/rosidl_python/commit/07ff2fcc47ef64bb600f1c54329d3bda22ac047e#r98895589

knorth55:
@sloretz kindly ping. can you review this PR?
#141 is a huge bug in humble and rolling, and this PR fixes the issue.

sillkjc:
It's rather insane we got to this far after Humble release and this is an issue. We're attempting to migrate our large code base from Galactic -> Humble and this has come up. We follow our own design pattern which does not include having messages in their own package due to the overhead required to maintain so many packages, as well as legacy. Then this gets sprung on us without any rep discussion, conversion warnings or understanding of how big an issue this is.

Sometimes you want messages and python in the one package. This or a similar fix really needs to be merged.

dgarcu:
Hi there, 

I recently have installed [audio_common](https://github.com/ros-drivers/audio_common/tree/ros2) package, [but I needed the assist](https://github.com/ros-drivers/audio_common/issues/227) of @knorth55 to do so. It is working just fine. Apparently the bug that prevented me to do it on my own is already fixed on his contributions. Just a friendly reminder! I am sure that it is worth the effort :smile: 

Greetings!

sloretz (commented):
Thanks for the PR!

It looks like this is working around an issue where `ament_python_install_package()` doesn't allow the same package name to be installed twice. I see why this would be important for packages that both generate interfaces and ship a python library. I don't think this is the right place for it. It exposes `ament_cmake_python`'s internal variable name `AMENT_CMAKE_PYTHON_INSTALL_INSTALLED_NAMES` to `rosidl_generator_py`, and once we add it here I don't see a path forward to eventually remove it.

For the time being I would recommend putting this code in the affected package itself instead of it calling `ament_python_install_package()` a second time.

I think the best option would be to update `ament_python_install_package()` to allow installing more files into an existing package.

For example,

```cmake
rosidl_generate_interfaces(...)

ament_python_install_package(... EXTEND_EXISTING)
```

knorth55:
@sloretz 

Thank you for your review.
I agree your idea and it is ideal, but it requires packages to change cmake.
Also, this issue stops many packages release, so merging this hot fix is also a good thing, IMO.
I will update `ament_python_install_package` near future, but I'm not sure when it will be.
I cannot make enough time to do that now.

MarcoMagriDev:
Any update on this?

clalancette:
> Any update on this?

I think we are waiting on an update to `ament_python_install_package`, which is the way forward that @sloretz pointed out.

knorth55:
@clalancette 
I don't get any reply about my suggestion, but this PR will not be merged as a hotfix?
I can work on to solve the issue in `ament_python_install_package`, but it will take time (on weekend project)...
https://github.com/ros2/rosidl_python/pull/187#issuecomment-1582106506

clalancette:
> I don't get any reply about my suggestion, but this PR will not be merged as a hotfix?

I don't think we should do that, no.  Workarounds like this tend to live forever once they are in.  Given that we have another way forward, I think we should do that.

knorth55:
@clalancette thanks. I see. I will work on the other way.

brta-jc:
At this point I don't really care if it stays forever, because it will allow us to move forward. It's been on an ongoing issue for a year now. This prevents us merging code that can run on ROS-industrial-CI pipelines on upstream Github. 
Please merge, revert later if necessary as this is moving way too slow.

RoBeau:
@sloretz could we get an update on this? 

- Can we get this hotfix applied yes or no?
- If no can we get support from a maintainer to work on a fix for `ament_python_install_package`
- Whats an approximate timeline for solution?

brta-jc:
@quarkytale for review?

rolker:
> Thanks for the PR!
> 
> It looks like this is working around an issue where `ament_python_install_package()` doesn't allow the same package name to be installed twice. I see why this would be important for packages that both generate interfaces and ship a python library. I don't think this is the right place for it. It exposes `ament_cmake_python`'s internal variable name `AMENT_CMAKE_PYTHON_INSTALL_INSTALLED_NAMES` to `rosidl_generator_py`, and once we add it here I don't see a path forward to eventually remove it.
> 
> For the time being I would recommend putting this code in the affected package itself instead of it calling `ament_python_install_package()` a second time.
> 
> I think the best option would be to update `ament_python_install_package()` to allow installing more files into an existing package.
> 
> For example,
> 
> ```cmake
> rosidl_generate_interfaces(...)
> 
> ament_python_install_package(... EXTEND_EXISTING)
> ```

Here's my attempt to solve this: https://github.com/ament/ament_cmake/pull/517
I would appreciate any help in improving the quality of the commit to make sure it get accepted and solves our problem.

clalancette:
See my comment in https://github.com/ros2/rosidl_python/issues/141#issuecomment-1986288774 .  I'm going to close this in favor of https://github.com/ament/ament_cmake/issues/514

Issue #192, opened 2023-01-31
Deserialization speedup proposal


almaslov:
## Feature request

Hello!
As already mentioned in other discussions python implementation has issues in serialization and deserialization performance(e.g. [here](https://discourse.ros.org/t/high-cpu-load-for-simple-python-nodes/28324/27)). I focus on deserialization here.
I've tried some fixes on `galactic` branch and they look very promising for me: depending on message structure and size it's about 3-20 times faster. x3 for simple messages, up to x20 for complex types with high nesting level and array fields.

#### Feature description
The main idea is to avoid pure python code in message deserialization. To achieve it, every message is split into C extension base class and pure python derived class. C extension class contains all message fields, so deserialization method initializes C structure members without calling python methods. On the other hand pure python class contains all dunder methods and assertions.

#### Implementation considerations
Here is an example of how generated files look like for simple message file [Example.msg](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/Example.msg):

[_example_base.c](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example_base.c) extension module defines base class containing members of native types in [lines 10-25](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example_base.c#L10).

Python module [_example.py](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example.py) is the same as in upstream except for base class([line 62](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example.py#L62)) and microoptimization of assertion in `__init__` method([line 95](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example.py#L95)).

Support extension module [_example_s.c](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example_s.c) now uses specific subtype of PyObject([line 147](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example_s.c#L147)), calls `__new__` instead of `__init__` to prevent pure python calls([line 170](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example_s.c#L170)), directly initializes C typed fields of message([lines 188, 191](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example_s.c#L188)) and creates nested collections ([line 193-212](https://github.com/almaslov/rosidl_generator_py_c_ext_example/blob/main/example_s.c#L193)).

[...]

PR #193, opened 2023-02-07
Using C extention base class for every message

almaslov:
Added C extension base class for every message. These base classes define native fields and corresponding slots(aka `struct PyMemberDef`). During deserialization they allow to call `__new__`(which is C function) instead of expensive `__init__`(which is pure Python) and initialize base class C fields instead of Python class members.

Details are in https://github.com/ros2/rosidl_python/issues/192


almaslov:
As far as I can understand errors in CI, my code depends on commit from `rosidl_parser` that is not tagged yet, and so not available in latest package from `http://repo.ros2.org/ubuntu/testing`. So CI will fill for all branches forked from current `rolling`.

clalancette:
> So CI will fill for all branches forked from current `rolling`.

Yeah, that's correct.  We need to do some releases, hopefully later this week.  We have other CI that we can run (by hand), so after this goes through review we can rely on that.

Voldivh:
Hi @almaslov, I was trying to test out your solution. However, I encountered an error during compilation once I switched to your branch. The error was:
```bash
make[2]: *** No rule to make target 'rosidl_generator_py/rosidl_generator_py/_rosidl_generator_py_bases.c', needed by 'CMakeFiles/rosidl_generator_py_custom__bases.dir/rosidl_generator_py/rosidl_generator_py/_rosidl_generator_py_bases.c.o'.  Stop.
```
I believe [this line](https://github.com/ros2/rosidl_python/compare/rolling...almaslov:rosidl_python:c-extention-base-class#diff-0da744c3127964a8907eb35fc0b5c04e32eccd515d8cf7fb839ec80ebb82fa59R197 ) is not generating the appropriate files or is not generating them on the proper path. Could you take a look at it?

almaslov:
Hello! Thank you for your time!
I've just tried to build `rosidl_generator_py` from scratch in fresh `ubuntu:jammy` docker container and had no problems. I've walked through this guide to setup environment for my build: https://docs.ros.org/en/rolling/Installation/Alternatives/Ubuntu-Development-Setup.html .

`_bases.c` file is generated in `build/rosidl_generator_py/rosidl_generator_py/rosidl_generator_py/_rosidl_generator_py_bases.c` as expected. 
I believe I had this issue while development and cleaning package build directory fixed it. If it doesn't help in your case, could you please describe your environment.

Voldivh:
Hi! After cleaning the environment it compiled without issue, I did notice that the problem happens when you try to compile on top of a previous build from other branches. 

The test I was trying to do was to publish a message (in particular a Image from the sensor_msgs) and measure the time it took from publishing the message until the subscriber received it. I believe that the message should pass through the serialization and deserealization of the message. I tried out that test with the rolling branch of this repository and the branch from your PR as well and both yielded basically the same amount of time. I thought I was going to be able to notice a difference in the latency of the messages when switching the branches. Could you perhaps point me in the right direction to test out your branch improvement? I'm not sure if perhaps the message was too small (The image was something around 200Kb) as to notice a difference in performance or if I should go with a custom message with several nested data structures. 
The way I was measuring the time was using the clock from the node, store that value in the header timestamp and checking the time again on the subscriber's callback.

almaslov:
As I can't see from same experiment on my local machine, deserialization takes about 150 microseconds and message transferring - 1150 microseconds. Deserialization on my feature branch takes about 130 microseconds, so optimization gain is less than 2%. I think the reason is that most of time is spent in memcpy'ing 200kb from C array to numpy array. I believe here is an area for optimization too, but it is out of current changes scope.

> or if I should go with a custom message with several nested data structures.

I think this case will be more demonstrative. For example array of `geographic_msgs/GeoPointStamped`(as in my project) or `TestArrayComplex` from [this issue](https://github.com/ros2/rmw_cyclonedds/issues/346#issue-1026807589)

Voldivh:
> I think this case will be more demonstrative. For example array of `geographic_msgs/GeoPointStamped`(as in my project) or `TestArrayComplex` from [this issue](https://github.com/ros2/rmw_cyclonedds/issues/346#issue-1026807589)

I like this idea, I'll go ahead and test it out with those messages and post the results here.
As for the PR, it seems to be compiling without any issues and is passing all the test on on my docker container with `ubuntu::focal`, I'll try it out on `ubuntu::jammy` as well but I don't think I would get any difference. We should wait for the next release and start another CI after that.

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-02-16/29927/1

Voldivh:
Hi @almaslov, sorry for the late response. I tested out the PR taking into account the message types and ubuntu distros as we talked before and I did notice an improvement as the data structures were more complex and had nested structures, as we expected it to happen. As for the building, everything went fine without any issues during compilation and testing.

Friendly ping to @clalancette , is the next rolling release out? I believe we can trigger a new CI in order to confirm everything is fine with merging this with upstream.

clalancette:
> Friendly ping to @clalancette , is the next rolling release out? I believe we can trigger a new CI in order to confirm everything is fine with merging this with upstream.

Yes, it is out.  I'll run the Rpr job again, but note that we'll also want to review this code and run full CI on it.

clalancette:

rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -156,8 +156,49 @@ PyObject * @('__'.join(type_.namespaces + [convert_camel_case_to_lower_case_unde
 @[  end if]@
 @[end for]@
 
+@# TODO(aamaslov) duplicated struct definition here
```
Voldivh, line 159: Can we remove the duplicate struct definition and use only one?

Voldivh (changes requested):
Left a some minor comments after the first pass of the changes. In overall I believe the PR is fine, however, I would like your opinion @wjwwood @sloretz .

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -103,6 +105,30 @@ class Metaclass_@(message.structure.namespaced_type.name)(type):
...
+        # There are two purposes for __slots__ field of message class:
+        #   1) __slots__ field is used in ROS as a list of message members.
+        #      Number of ROS packages rely on its value.
+        #   2) defining __slots__ in Python class prevents adding __dict__
+        #      to class instances. This behvior is desirable.
+        # If #1 defined in message class body, then base class member descriptors are
+        # overriden and no longer can be used. On the other side, any __slots__ value
+        # should be defined to guaratee #2.
+        # So we define empty list in message class body and modify it after type object
+        # constructed. This is neither prohibited, nor encouraged by CPython documentation.
```
sloretz, line 121: If I understand correctly, normally `__slots__` requires subclasses to only define new slots:

> ["The action of a __slots__ declaration is not limited to the class where it is defined. __slots__ declared in parents are available in child classes. However, child subclasses will get a and __weakref__ unless they also define __slots__ (which should only contain names of any additional slots)."](https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots)

This defines an empty `__slots__` attribute on the subclass prior to calling `type.__new__`, which is where cPython makes the decision to add `__dict__` or not, and where the quoted note seems to matter. It then adds to `__slots__`  afterwards with the assumption that cPython won't use it for any other purpose, but it allows users to get a list of the members of a message.

I'm hesitant because it seems not very well defined when cPython will stop using `__slots__`.

An alternative option would be to make a different API for retrieving the members of a message. That would require downstream users to change their code in Rolling, but would give us freedom to put `__slots__` in base classes. 

Another alternative would be to implement the entire class in C so that there's only one class.

Maybe yet another option would be to do monkey patching instead of a metaclass. Say there was a `class Example:`  that had an `Example.__new__` method that returned an instance of `ExampleBase`. The first time it ran it would monkey patch `ExampleBase` with the methods from `Example`.

Out of all the options, my favorite is making a new API for retrieving members of the message. @clalancette What do you think? Pinging you because it would change a (maybe unintentionally) public API without a good way to tick/tock it.

sloretz (commented):
I think the general idea is a good one. It will take me a bit to give it a thorough review though.

rosidl_generator_py/resource/_import_type.c.em
```diff
@@ -0,0 +1,32 @@
+@{
+module_name, type_name = full_type_name.rsplit(".", maxsplit=1)
+ensure_is_type = locals().get('ensure_is_type', False)
+}@
+@
+static PyObject * @(func_name)()
+{
+  // TODO(almaslov) proper deinitialization
```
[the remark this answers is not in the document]
almaslov: Sorry for late response.
I’ve researched a little about best practice for this case, and in short, I think it will reqire changing public API.
First let me explain what is this comment about. Deserialization function should import a message class to be able to construct its instance. In current upstream solution it is imported on every call. So if large message contains 1000 submessages, than import machinery is called 1000 times too. And thought repeated calls of module import is relatively cheap, in total 1000 imports are quite expensive. `_import_type.c.em` is a naive implementation of cache which stores cached value in static C variable. First problem here is that imported type objects are not decref’ed if support module unimported and will be garbage collected only on interpreter exit. The second problem potential issues with python subinterpreters ([some info here](https://peps.python.org/pep-0554/)).

I think these imported type objects should be stored in some state-object to be finalized later. It could be implemented in some ways
* Python3 has recommende way to store global state - PyModule_GetState, which is described here: https://peps.python.org/pep-3121/. 
* rclpy bindings could pass some special state-object and manage its lifetime to implement more complicated logic. 

As I can see, additional parameter with state-object required by `convert_to_py` and `convert_from_py` C methods in any way.
What do you think about it?

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -202,14 +228,9 @@ class @(message.structure.namespaced_type.name)(metaclass=Metaclass_@(message.st
...
-@[for member in message.structure.members]@
-@[  if len(message.structure.members) == 1 and member.name == EMPTY_STRUCTURE_REQUIRED_MEMBER_NAME]@
-@[    continue]@
-@[  end if]@
-        '_@(member.name)',
-@[end for]@
-    ]
+    # This field is modified after class creation.
+    # See the comment to Metaclass_@(message.structure.namespaced_type.name).__new__
+    __slots__ = []
```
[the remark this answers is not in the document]
almaslov: Description from `sloretz` is very accurate.

rosidl_generator_py/resource/_import_type.c.em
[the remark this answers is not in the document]
Voldivh: Sorry for the late response as well,  I now understand a little better the use of `_import_type.c.em` and the issue at hand. The general idea of the module seems fine to me. In regards to the issue at hand,  after reading a bit about the solution you propose I like the  idea in order to avoid memory leaks. The only thing that worries me is the modification of the public API, Do you have a general idea of how much changes this would imply?

Another possible solution could be to manually handle the decref of the imported type objects by using `Py_DECREF()` at the moment of unloading the module. Do you think it could work? Let me know what you think.

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-03-16/30432/1

rosidl_generator_py/resource/_msg_support.c.em
almaslov: Struct is moved to `@(package_name)_decl.h` header file.

rosidl_generator_py/resource/_msg.py.em
clalancette: Yeah, I have to agree with @sloretz ; while this might currently work for `__slots__`, it seems like something that could break in future versions of Python.   I don't think we should rely on that behavior here.

> Out of all the options, my favorite is making a new API for retrieving members of the message. @clalancette What do you think? Pinging you because it would change a (maybe unintentionally) public API without a good way to tick/tock it.

I'm curious what you have in mind here?  We already have `property`/`setter` methods for the real class that the user interacts with.  Are you suggesting adding a new API to the Metaclass?  And what would that be?

rosidl_generator_py/resource/_import_type.c.em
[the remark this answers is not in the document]
mjcarroll: @almaslov and I discussed this a bit offline.  I can't think of any instance where we are using sub-interpreters, so this should _likely_ work.  I think that we should add a comment in the code to indicate that this is unsupported in the case of sub-interpreters.  

I think that the likely scenario is that you would end up with a double-free in that case (though that is a guess), and we should save someone some time in the future if they run into this unsupported use case.

rosidl_generator_py/resource/_msg.py.em
sloretz: > I'm curious what you have in mind here? We already have property/setter methods for the real class that the user interacts with. Are you suggesting adding a new API to the Metaclass? And what would that be?

I was suggesting a new API to get the message fields, followed by replacing any downstream use of `__slots__` with that API. Then we no longer need to modify `__slots__` here as there's already a supported way to get message fields.

Turns out that API already exists 🎉 , so I think we can get rid of modifying `__slots__` altogether.

```python3
>>> from sensor_msgs.msg import JointState
>>> JointState.get_fields_and_field_types()
{'header': 'std_msgs/Header', 'name': 'sequence<string>', 'position': 'sequence<double>', 'velocity': 'sequence<double>', 'effort': 'sequence<double>'}
```

sloretz: I'm not saying we get rid of `__slots__`. I'm saying we don't need to make sure the subclass contains all the message fields. They're already in the `__slots__` of the base class.

clalancette: > I'm not saying we get rid of __slots__. I'm saying we don't need to make sure the subclass contains all the message fields. They're already in the __slots__ of the base class.

Yes!  Thank you for the clear explanation in person, that helped.  Now I understand and I agree.

Voldivh: Thanks @almaslov! Just as a comment, on [#194](https://github.com/ros2/rosidl_python/pull/194#issuecomment-1490900345) we encountered the issue of other packages using `__slots__` instead of the method `get_fields_and_field_types()`. I've been working on some PRs on those packages to solve that issue. All the information regarding which packages have been solved or not is on the conversation of #194.

EsipovPA:
Hello everyone!

I've took the liberty of updating this PR.
I've notices that there were some conflicts with the **rolling** branch.
Looks like these are done with.

Is it OK to request a new review?

EsipovPA:
@cottsay @sloretz 
Hello guys! 

it seems like tests have finished successfully.
May I ask you to take a look at this PR, when you have a time?

Just a gentle reminder :slightly_smiling_face:

EsipovPA:
Hello guys! I guess, this PR needs some rebaseing again. Is there any chance of moving it forward after that? 

@InvincibleRMC @sloretz Is there a chance to review this again? This PR have been in review for quite some time now, but I'm willing to get it to a finish line in the end sometime. :)

Seems like a valuable feature to me after all, but it needs some attention from maintainers team to get into `rolling`.

InvincibleRMC:
Yeah I can give this a review. Can you fix the current merge conflicts? Do you have some other repo scripts were you ran your performance comparison? I'm curious if the baseline has changed at all.

EsipovPA:
@InvincibleRMC I'll rebase this and look into what's left from testing environment for this PR. I think I'll be able to find my old scripts. If not, I'll write new. They were not that complicated in the end, as I recall.

EsipovPA:
The rebase is taking some time. Still working on it. Sorry for the delay. `.em` templates take some time to understand) Forgot everything since the last time.

upd (09.18.26)

I've rebased the source branch and it does  compile successfully, but I get error and a core du ()mp when comparing types in the beginning of the message serialization. Working on it.

PR #194, opened 2023-03-21
Hides the assertions that checks the data types of the message fields.

Voldivh:
## Context
This PR is related to the issue described in [this discussion](https://discourse.ros.org/t/high-cpu-load-for-simple-python-nodes/28324). After doing some benchmarks on simple publishers and subscribers on rclpy using different types of messages (Ex: Strings, Ints, Arrays, Custom messages with nested structures) and analyzing their behavior using flamegraphs, we noticed that a lot of the busy time of the nodes was being used during the creation of the messages object. Taking that into account, a possible area of improvement for this issue is by hiding all the assertions made on the fields of the messages under a user controlled flag.

## Description
 In summary, this PR hides all the assertions made at the time of construction of messages in rclpy. In order to give the user control whether to do the assertions (and lose performance due to this) or to not do the assertions, all the message constructors have  a new argument called `check_fields` which default value is `False`. This way all the pre-existing code would notice this improvement without the need of doing any modification.

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -372,7 +372,7 @@ if isinstance(type_, AbstractNestedType):
         typename.pop()
         typename.append(self.__class__.__name__)
         args = []
-        for s, t in zip(self.get_fields_and_field_types().keys, self.SLOT_TYPES):
+        for s, t in zip(self.get_fields_and_field_types().keys(), self.SLOT_TYPES):
             field = getattr(self, s)
             fieldstr = repr(field)
             # We use Python array type for fields that can be directly stored
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -253,7 +253,6 @@ string@
 @[  end if]@
 ',
 @[end for]@
-        '_check_fields': 'boolean',
     }
 
     SLOT_TYPES = (
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -289,7 +289,6 @@ if isinstance(type_, AbstractNestedType):
 @[  end if]@
 ,  # noqa: E501
 @[end for]@
-        rosidl_parser.definition.BasicType('boolean'),  # noqa: E501
     )
 
     def __init__(self, **kwargs):
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -371,7 +371,7 @@ if isinstance(type_, AbstractNestedType):
         typename.pop()
         typename.append(self.__class__.__name__)
         args = []
-        for s, t in self.get_fields_and_field_types.items():
+        for s, t in self.get_fields_and_field_types().items():
             field = getattr(self, s)
             fieldstr = repr(field)
             # We use Python array type for fields that can be directly stored
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -293,7 +293,7 @@ if isinstance(type_, AbstractNestedType):
     )
 
     def __init__(self, check_fields = False, **kwargs):
-        self._check_fields = check_fields or ge
+        self._check_fields = check_fields
         if self._check_fields:
             assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
                 'Invalid arguments passed to constructor: %s' % \
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -255,6 +255,8 @@ string@
 @[end for]@
     }
 
+    # This attribute is used to store an rosidl_parser.definition variable
+    # related to the data type of each of the components the message.
     SLOT_TYPES = (
 @[for member in message.structure.members]@
 @[  if len(message.structure.members) == 1 and member.name == EMPTY_STRUCTURE_REQUIRED_MEMBER_NAME]@
@@ -370,7 +372,7 @@ if isinstance(type_, AbstractNestedType):
         typename.pop()
         typename.append(self.__class__.__name__)
         args = []
-        for s, t in self.get_fields_and_field_types().items():
+        for s, t in zip(self.get_fields_and_field_types().keys, self.SLOT_TYPES):
             field = getattr(self, s)
             fieldstr = repr(field)
             # We use Python array type for fields that can be directly stored
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -209,6 +209,7 @@ class @(message.structure.namespaced_type.name)(metaclass=Metaclass_@(message.st
 @[  end if]@
         '_@(member.name)',
 @[end for]@
+        '_check_fields',
     ]
 
     _fields_and_field_types = {
@@ -252,6 +253,7 @@ string@
 @[  end if]@
 ',
 @[end for]@
+        'check_fields': 'boolean',
     }
 
     SLOT_TYPES = (
@@ -290,10 +292,12 @@ if isinstance(type_, AbstractNestedType):
 @[end for]@
     )
 
-    def __init__(self, **kwargs):
-        assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
-            'Invalid arguments passed to constructor: %s' % \
-            ', '.join(sorted(k for k in kwargs.keys() if '_' + k not in self.__slots__))
+    def __init__(self, check_fields = False, **kwargs):
+        self._check_fields = check_fields or ge
+        if self._check_fields:
+            assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
+                'Invalid arguments passed to constructor: %s' % \
+                ', '.join(sorted(k for k in kwargs.keys() if '_' + k not in self.__slots__))
 @[for member in message.structure.members]@
 @[  if len(message.structure.members) == 1 and member.name == EMPTY_STRUCTURE_REQUIRED_MEMBER_NAME]@
 @[    continue]@
@@ -378,7 +382,8 @@ if isinstance(type_, AbstractNestedType):
                 if len(field) == 0:
                     fieldstr = '[]'
                 else:
-                    assert fieldstr.startswith('array(')
+                    if self._check_fields:
+                        assert fieldstr.startswith('array(')
                     prefix = "array('X', "
                     suffix = ')'
                     fieldstr = fieldstr[len(prefix):-len(suffix)]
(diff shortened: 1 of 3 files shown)


rosidl_generator_py/resource/_msg.py.em
```diff
@@ -290,10 +292,12 @@ if isinstance(type_, AbstractNestedType):
 @[end for]@
     )
 
-    def __init__(self, **kwargs):
-        assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
-            'Invalid arguments passed to constructor: %s' % \
-            ', '.join(sorted(k for k in kwargs.keys() if '_' + k not in self.__slots__))
+    def __init__(self, check_fields = False, **kwargs):
+        self._check_fields = check_fields
```
Voldivh, line 296: As part of the discussion of this PR I want to comment that my initial solution, besides having a new argument on the constructor of the messages, also retrieved a flag that could also be used to enable the assertions. The way I was doing it was basically the following:

```suggestion
    from os import getenv
    def __init__(self, check_fields = False, **kwargs):
        self._check_fields = check_fields or getenv("ROS_PYTHON_CHECK_FIELDS")
```
That way the member `self._check_fields` could be configured via the constructor of each message or just by setting the an environment variable to `True`.

I removed this solution because after testing it out, the time it took to retrieve the environment value basically made the solution's improvement almost non relevant. I believe that incorporating a way for the user to select to have the fields checked or not by just configuring a single variable for the whole node/project could be useful for the users. I would like to hear your thoughts on this @sloretz @clalancette

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -87,6 +87,10 @@ for member in message.structure.members:
 @[end if]@
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
+from os import getenv
+
+ros_python_check_fields = getenv("ROS_PYTHON_CHECK_FIELDS")
```
clalancette, line 92: I think it is worthwhile to put a comment here saying that you are doing this at the module level since it was too slow at the per-object level.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -288,12 +294,15 @@ if isinstance(type_, AbstractNestedType):
...
 @[end for]@
+        rosidl_parser.definition.BasicType('boolean'),  # noqa: E501
     )
 
-    def __init__(self, **kwargs):
-        assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
-            'Invalid arguments passed to constructor: %s' % \
-            ', '.join(sorted(k for k in kwargs.keys() if '_' + k not in self.__slots__))
+    def __init__(self, check_fields=False, **kwargs):
+        self._check_fields = check_fields
```
clalancette, line 301: If you change this to:

```suggestion
        self._check_fields = check_fields or ros_python_check_fields
```

Then you only need to do `self._check_fields` below.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -428,28 +438,28 @@ if member.name in dict(inspect.getmembers(builtins)).keys():
...
+            if isinstance(value, array.array):
+                assert value.typecode == '@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['type_code'])', \
+                    "The '@(member.name)' array.array() must have the type code of '@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['type_code'])'"
+@[          if isinstance(member.type, BoundedSequence)]@
+                assert len(value) <= @(member.type.maximum_size), \
+                    "The '@(member.name)' array.array() must have a size <= @(member.type.maximum_size)"
+@[          end if]@
+                self._@(member.name) = value
+                return
+@[        end if]@
```
clalancette, line 461: I'm not totally sure, but I think the `empy` if statements are now over-indented.  That is, they are all inside of the `@[for member in message.structure.members]@` loop starting at line 418, so I think they should only be indented by two spaces.

That will also make it easier to review this PR, as the only change will be from `if __debug__` to `if self._check_fields`.

clalancette (changes requested):
The general idea looks good to me.  I've left a few things that I think should be fixed.  Once that is done, I think we can run CI on it.

rosidl_generator_py/resource/_msg.py.em
Voldivh: Done. Take a look and let me know if the comment is fine.

rosidl_generator_py/resource/_msg.py.em
Voldivh: Thanks and done.

rosidl_generator_py/resource/_msg.py.em
Voldivh: Yeah, the `empy` statements were over-indented. I already fixed it and I think it is easier to review that part as well.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -288,12 +297,15 @@ if isinstance(type_, AbstractNestedType):
...
 @[end for]@
+        rosidl_parser.definition.BasicType('boolean'),  # noqa: E501
     )
 
-    def __init__(self, **kwargs):
-        assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
-            'Invalid arguments passed to constructor: %s' % \
-            ', '.join(sorted(k for k in kwargs.keys() if '_' + k not in self.__slots__))
+    def __init__(self, check_fields=False, **kwargs):
+        self._check_fields = check_fields or ros_python_check_fields
```
clalancette, line 304: After playing with this for a bit, I actually think we should change up the logic here.

There are currently two problems that I see:
1. Doing `check_fields or ros_python_check_fields` isn't actually correct; if the `ROS_PYTHON_CHECK_FIELDS` environment variable is not defined, `ros_python_check_fields` evaluates to `None`, and if it is defined, it evaluates to a string.
2. If the user explicitly passes `check_fields=False`, then we probably want to honor that, *even if ROS_PYTHON_CHECK_FIELDS was also set to true*.

To solve both of these, I think we need to do the following:

```suggestion
    def __init__(self, **kwargs):
        if 'check_fields' in kwargs:
            self._check_fields = kwargs['check_fields']
        else:
            self._check_fields = ros_python_check_fields is not None and ros_python_check_fields == "1"
```

clalancette (changes requested):

rosidl_generator_py/resource/_msg.py.em
clalancette: > I think this would make more sense than to leave it as none. Regarding the value the user sets to the variable, I'm a little troubled since the can basically set the variable to anything he wants right? Wouldn't it makes sense to add a check up during the lookup of the variable?
> ```

Yes, they can set it to anything.  But my thinking is that maybe in the future, we want to have more fine-grained checking.  Like if you set the environment variable to "1", it only checks simple types, but if you set it to "2", it also checks array elements.  We don't need to define all of the semantics here, but if we make sure that this is a known value, we can easily expand it in the future.

So with that in mind, I'll suggest we combine your approach and my approach.  Basically when looking up the variable with getenv, we do:

```
from os import getenv
ros_python_check_fields = getenv("ROS_PYTHON_CHECK_FIELDS", default='') 
```

And then when we use it, we do:
```
        if 'check_fields' in kwargs:
            self._check_fields = kwargs['check_fields']
        else:
            self._check_fields = ros_python_check_fields == "1"
```

Does that make sense to you?

Voldivh: Ok, yeah it makes sense. I like the idea of  expanding it in the future if needed/wanted.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -87,6 +87,13 @@ for member in message.structure.members:
 @[end if]@
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
+# This is being done at the module level and not on the instance level to avoid looking
+# for the same variable multiple times on each instance. This variable is not supposed to
+# change during runtime so it makes sense to only look for it once.
+from os import getenv
+
+ros_python_check_fields = getenv("ROS_PYTHON_CHECK_FIELDS", default='')
```
clalancette, line 95:
```suggestion
ros_python_check_fields = getenv('ROS_PYTHON_CHECK_FIELDS', default='')
```

clalancette: Also, when I run the tests on this, flake8 says that the imports are in the wrong order.  So this probably needs to move above somehow.

clalancette (changes requested):
A few more things to fix for flake8.

rosidl_generator_py/resource/_msg.py.em
Voldivh: That's weird, when I run the test it isn't showing me that. It did before and I moved it up and stop appearing the error. I'll check.

clalancette: You have to run one of the tests for the generated output.  I did:

```
colcon test --event-handlers console_direct+ --packages-select visualization_msgs
```

Voldivh: I see the problem now. I was running the test for only this package instead of the whole build. The problem is that for actions and services the `_msg.py.em` template is being called multiple times and the import appears multiple times as well. I think a solution would be to move that import up one module level and import it on the `_idl.py.em` module. Let me know what you think.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18311/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12833/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18977/)

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -252,6 +253,7 @@ string@
 @[  end if]@
 ',
 @[end for]@
+        'check_fields': 'boolean',
```
clalancette, line 256:
```suggestion
        '_check_fields': 'boolean',
```

clalancette (changes requested):

rosidl_generator_py/resource/_msg.py.em
Voldivh: Thanks and done.

Voldivh:
After a little research and an internal discussion. We found out that the reason why some packages such as `rqt_py_common` and `ros2action` are failing CI is because some API's are using the `__slots__` attribute to retrieve the message's fields and field types instead of using the method `get_fields_and_field_types()`. I created a [PR](https://github.com/ros2/rosidl_runtime_py/pull/23) that should solve the problem.

clalancette:
So with the fixes elsewhere, I think we are getting closer here.

However, I think we have another problem in this repository.  In particular, we want the `__repr__` of each message to only represent the fields of the message, and it currently uses the `__slots__`:

https://github.com/ros2/rosidl_python/blob/e6d72487a05b64792e437406058bd0f5740640c5/rosidl_generator_py/resource/_msg.py.em#L362-L386

What I'll suggest here is to just switch that `for s, t in zip(self.__slots__, self.SLOT_TYPES):` to `for s, t in self._fields_and_field_types.items():`  at https://github.com/ros2/rosidl_python/blob/e6d72487a05b64792e437406058bd0f5740640c5/rosidl_generator_py/resource/_msg.py.em#L367 .  You'll also have to update https://github.com/ros2/rosidl_python/blob/e6d72487a05b64792e437406058bd0f5740640c5/rosidl_generator_py/resource/_msg.py.em#L385 to not strip off the first character, which was `_` when we were using slots.

Voldivh:
I see your point. I'll get right down to it. However, I think that we might actually encounter similar problems in other packages, is it worth that I do a little search and modifications or should we just leave it as it is as long as CI passes? @clalancette

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -364,7 +370,7 @@ if isinstance(type_, AbstractNestedType):
         typename.pop()
         typename.append(self.__class__.__name__)
         args = []
-        for s, t in zip(self.__slots__, self.SLOT_TYPES):
+        for s, t in self.get_fields_and_field_types().items():
```
clalancette, line 373: This needs to be fixed to use `zip(self.get_fields_and_field_types().keys, self.SLOT_TYPES)`

clalancette (changes requested):
Besides the comment inline, we still have to fix the `__repr__` method in here to generate the correct output (using the `zip` method again).

rosidl_generator_py/resource/_msg.py.em
Voldivh: Oh you are right.

Voldivh: Done.

clalancette:
So I did a bit more digging into how `__slots__` is used throughout the codebase.

To do this, I fetched *all* of the code that is currently released into rolling with the following:
```
$ mkdir ros2_rolling
$ cd ros2_rolling
$ rosinstall_generator ALL --rosdistro rolling --deps | vcs import
```

(this takes a while).  I then grepped through all of the code to find uses of `__slots__`:

```
$ grep -rI "__slots__" *
```

There are a total of 122 uses of `__slots__` throughout the codebase.  Many of them are just regular Python code, just using `__slots__` for their own purposes.

However, I did find a number of uses of `__slots__` from ROS messages.  Most of the usage of these is going to break in subtle ways if we put this PR in without doing something about it.  Here is the list I found (ignoring `rosidl_runtime_py` since https://github.com/ros2/rosidl_runtime_py/pull/23 should fix that):

* [rosbridge_suite](https://github.com/RobotWebTools/rosbridge_suite/blob/ef0e4172667068b4c582a7277fe0b3071fa06aaf/rosapi/src/rosapi/objectutils.py#L144)
* [rospy_message_converter](https://github.com/DFKI-NI/rospy_message_converter/blob/bcce27e2af48810e33bbb7e3368b65763e49d08b/rclpy_message_converter/rclpy_message_converter/message_converter.py#L217)
* [rqt_py_common](https://github.com/ros-visualization/rqt/blob/95e76836b24880b8d8fcc952d7da93da260e2b9f/rqt_py_common/src/rqt_py_common/topic_dict.py#L71)
* [rqt_bag](https://github.com/ros-visualization/rqt_bag/blob/2f6fbd558baa8e8038f4d11bb20e27c75537d39c/rqt_bag_plugins/src/rqt_bag_plugins/plot_view.py#L437)
* [rqt_plot](https://github.com/ros-visualization/rqt_plot/blob/726c375389ed2b91313742008a5cf04bd0e155b7/src/rqt_plot/plot_widget.py#L117)
* [rqt_publisher](https://github.com/ros-visualization/rqt_publisher/blob/65805511661c086df7282bcc9e37f79f58ec3068/src/rqt_publisher/publisher.py#L181)
* [rqt_robot_monitor](https://github.com/ros-visualization/rqt_robot_monitor/blob/43b9126c1fb8af37a5ef49c636283bf40a23f614/src/rqt_robot_monitor/robot_monitor_bag_plugin.py#L70)

I think we should attempt to fix these before we put this in, otherwise things will start failing and it will be hard to tell why.  @Voldivh can you start taking a look at this, particularly `rqt_py_common`, `rqt_bag`, `rqt_plot`, and `rqt_publisher`?

Voldivh:
> I think we should attempt to fix these before we put this in, otherwise things will start failing and it will be hard to tell why.  @Voldivh can you start taking a look at this, particularly `rqt_py_common`, `rqt_bag`, `rqt_plot`, and `rqt_publisher`?

Yeah, after looking a little bit to the packages you mentioned, I can see that they will require some fixing. I'll start to work on the ones you commented.

Voldivh:
Just finished created the PR's that solves the issues with the packages: 

- [rqt_py_common](https://github.com/ros-visualization/rqt/pull/289)
- [rqt_bag](https://github.com/ros-visualization/rqt_bag/pull/138)
- [rqt_plot](https://github.com/ros-visualization/rqt_plot/pull/87)
- [rqt_publisher](https://github.com/ros-visualization/rqt_publisher/pull/38).

clalancette:
OK, so we now have the 4 in-core uses of slots fixed.

I think it is worthwhile to open new issues or PRs for `rosbridge_suite`, `rospy_message_converter`, and `rqt_robot_monitor`, telling them that this use is changing and how to fix it.  Once that is done, we can go ahead and run CI here.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18442/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12988/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19128/)

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -364,7 +372,7 @@ if isinstance(type_, AbstractNestedType):
         typename.pop()
         typename.append(self.__class__.__name__)
         args = []
-        for s, t in zip(self.__slots__, self.SLOT_TYPES):
+        for s, t in zip(self.get_fields_and_field_types().keys, self.SLOT_TYPES):
```
clalancette, line 375:
```suggestion
        for s, t in zip(self.get_fields_and_field_types().keys(), self.SLOT_TYPES):
```

clalancette (changes requested):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18458/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13004/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19150/)

clalancette:
@Voldivh OK, this is looking good to me with CI.  Can you please open up a PR to https://github.com/ros2/ros2_documentation/blob/rolling/source/Releases/Release-Iron-Irwini.rst , saying that we changed this behavior?  Once that is ready, I'll merge that and this one together.

PR #195, opened 2023-05-16
Install compiled libraries only to 'lib'

cottsay:
The current behavior results in duplicate libraries installed to both the Python module in libdir and also into the global lib directory for the prefix.

The former seems unnecessary, so we should install the libraries only to lib.

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -146,17 +146,13 @@ macro(set_properties _build_type)
   set_target_properties(${_target_name} PROPERTIES
     COMPILE_OPTIONS "${_extension_compile_flags}"
     PREFIX ""
-    LIBRARY_OUTPUT_DIRECTORY${_build_type} ${_output_path}
-    RUNTIME_OUTPUT_DIRECTORY${_build_type} ${_output_path}
     OUTPUT_NAME "${PROJECT_NAME}_s__${_typesupport_impl}${PythonExtra_EXTENSION_SUFFIX}"
     SUFFIX "${PythonExtra_EXTENSION_EXTENSION}")
 endmacro()
 
 macro(set_lib_properties _build_type)
   set_target_properties(${_target_name_lib} PROPERTIES
-    COMPILE_OPTIONS "${_extension_compile_flags}"
-    LIBRARY_OUTPUT_DIRECTORY${_build_type} ${_output_path}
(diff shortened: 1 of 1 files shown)


cottsay:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18753/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13279/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19494/)

clalancette (commented):
So I'm not at all sure about this, but while looking this over I noticed something.  That is the fact that if we remove the lines you are proposing to remove, then there is no reason to pass `_build_type` into these two macros anymore.

In turn, that leads me to look at where the macros are being called.  They are called once with an "empty" (I assume "Release") build type for all platforms.  And then, only on WIN32, they are called once for each build type.  I don't exactly know what that means.

Looking at the history of this code, it was added in 2016 (via 1d7b4eb15a0fae579ec4e02f1345156886235200 before we used PRs) and 2018 (via #10).  So this code has been there quite a while.

I think we need to do some more extensive testing on this besides the normal CI.  In particular, we should also test this out with Windows Debug to see what happens.  If we can get it to work there, then I'm fine with the change, but we should then also remove the `build_type` argument from both macros (and maybe get rid of the macros completely).

brta-jc:
Bump. Any updates or plan for this?

cottsay:
> Bump. Any updates or plan for this?

Yep:

> I think we need to do some more extensive testing on this

By all means, if you're interested in seeing this merged, please test it and provide feedback.

clalancette:
I've updated this to the latest, and also removed what I think are now unnecessary macros.  Here's a large set of CI to see what happens:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19755/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14257/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20480/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20481/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/475/)
* Clang [Build Status](http://ci.ros2.org/job/ci_linux_clang_libcxx/38/)

clalancette:
All right, the extensive CI is actually really happy with this.  So I'm going to go ahead and merge this in, since it seems to be an improvement, and doesn't seem to have downsides.  After a few days in source, I'll likely also do a release of this just to be sure the buildfarm is happy with it as well.

clalancette (approved):

Issue #196, opened 2023-06-13
Python docstring for message documentation


guillaumeautran:
## Enhancement Request

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Source
- Version or commit hash:
  - git branch: Humble
- DDS implementation:
  - N/A
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
When looking at a message class documentation in Python, the message comments don't really show making it more complicated to understand the message. 

For example take the [statistics_msgs/MetricsMessage](https://github.com/ros2/rcl_interfaces/blob/rolling/statistics_msgs/msg/MetricsMessage.msg), the source file has lots of good documentation for the message itself as well as for each of the fields contained in the message. 
When looking at the python documentation for the message, we see:
```
Help on class MetricsMessage in module statistics_msgs.msg._metrics_message:

class MetricsMessage(builtins.object)
 |  MetricsMessage(**kwargs)
 |  
 |  Message class 'MetricsMessage'.
 |  
 |  Methods defined here:
 |  
 |  __eq__(self, other)
 |      Return self==value.
 |  
 |  __init__(self, **kwargs)
 |      Initialize self.  See help(type(self)) for accurate signature.
 |  
 |  __repr__(self)
 |      Return repr(self).
 |  
 |  ----------------------------------------------------------------------
 |  Class methods defined here:
 |  
 |  get_fields_and_field_types() from statistics_msgs.msg._metrics_message.Metaclass_MetricsMessage
 |  
 |  ----------------------------------------------------------------------
 |  Data descriptors defined here:
 |  
 |  measurement_source_name
 |      Message field 'measurement_source_name'.
 |  
 |  metrics_source
 |      Message field 'metrics_source'.
 |  
 |  statistics
 |      Message field 'statistics'.
 |  
 |  unit
 |      Message field 'unit'.
 |  
 |  window_start
 |      Message field 'window_start'.
 |  
 |  window_stop
 |      Message field 'window_stop'.
 |  
 |  ----------------------------------------------------------------------
 |  Data and other attributes defined here:
 |  
 |  SLOT_TYPES = (<rosidl_parser.definition.UnboundedString object>, <rosi...
 |  
 |  __hash__ = None
```
All of this without any of the message creator help data. 
#### Expected behavior
Instead, we could build in the documentation comments from the original message documentation into the python class docstring making it better for developers to access the documentation.
A potential outcome for this message type: 
```
[...]

PR #197, opened 2023-06-13
Edited the template to provide better docstring for message definition

guillaumeautran:
issue: #196

Transcribed the original message comments in the python docstring


guillaumeautran:
@sloretz Any feedback you can provide on this work?

Issue #198, opened 2023-07-14
Missing dependency on `ament_cmake_flake8`?


130s:
## Issue report

**Required Info:**

- Operating System: humble
- Installation type: Bin
- Version or commit hash:
   ```
   # ros2 pkg xml rosidl_generator_py                                                  
   <package format="3">
     <name>rosidl_generator_py</name>
     <version>0.18.0</version>
   ```

#### Steps to reproduce issue
Buildfarm output https://build.ros2.org/job/Rpr__smach__ubuntu_jammy_amd64/2/console for https://github.com/ros/executive_smach/pull/112 where a dependency on `ament_cmake_flake8` is removed in a downstream package (`smach_msgs`) so that none of the packages in the repo in the PR depends on nor call `ament_cmake_flake8` any more.

<details><summary>(I haven't figured out exactly what command was called when the reported error occured. My guess is either some sort of `rosidl` command was called, or `colcon build`</summary>

```
:
17:32:59 -- Found PythonExtra: .so  
17:32:59 CMake Error at /opt/ros/rolling/share/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake:335 (find_package):
17:32:59   By not providing "Findament_cmake_flake8.cmake" in CMAKE_MODULE_PATH this
17:32:59   project has asked CMake to find a package configuration file provided by
17:32:59   "ament_cmake_flake8", but CMake did not find one.
17:32:59 
17:32:59   Could not find a package configuration file provided by
17:32:59   "ament_cmake_flake8" with any of the following names:
17:32:59 
17:32:59     ament_cmake_flake8Config.cmake
17:32:59     ament_cmake_flake8-config.cmake
17:32:59 
17:32:59   Add the installation prefix of "ament_cmake_flake8" to CMAKE_PREFIX_PATH or
17:32:59   set "ament_cmake_flake8_DIR" to a directory containing one of the above
17:32:59   files.  If "ament_cmake_flake8" provides a separate development package or
```
</details>

#### Expected behavior

In a downstream interface package, no error during `cmake` regarding `ament_cmake_flake8Config.cmake / ament_cmake_flake8-config.cmake` should be seen while building rosidl package, even when the downstream interface package does NOT explicitly depend on `ament_cmake_flake8`.

#### Actual behavior

The error posted in "Steps to reproduce issue" section occurs.

130s:
[...]

PR #199, opened 2023-07-14
Fix: Missing dependency that causes cmake error in downstream (resolves https://github.com/ros2/rosidl_python/issues/198)

130s:
# Issue aimed at
- Closes https://github.com/ros2/rosidl_python/issues/198

# Changes
- Add a missing package in `package.xml`.
   - Even if that the pkg in question is only required conditionally https://github.com/ros2/rosidl_python/issues/198#issuecomment-1636500998, it makes more sense to me in the upstream (here) to install it than making it a job for downstream pkgs.
   - That said, I have little idea if the implementation I'm suggesting is the best/reasonable. Open for advises.

### DIFF
--- a/rosidl_generator_py/package.xml
+++ b/rosidl_generator_py/package.xml
@@ -29,16 +29,16 @@
   <buildtool_export_depend>rosidl_typesupport_c</buildtool_export_depend>
   <buildtool_export_depend>rosidl_typesupport_interface</buildtool_export_depend>
 
-  <!-- BEGIN: Following 'exec_depend's are likely only needed in test in downstream,
-       so ideally "test_export_depend" (similar to 'build_export_depend') are reasonable.
-       With such a feature unexistent yet, they are 'exec_depend'-ed.
+  <!-- TODO: The following 'exec_depend's are only needed in downstream tests,
+       so ideally a 'test_export_depend' (similar to 'build_export_depend') would be used.
+       Since that feature doesn't exist yet, they are 'exec_depend'-ed.
        See https://github.com/ros2/rosidl_python/pull/199 -->
   <exec_depend>ament_cmake_cppcheck</exec_depend>
   <exec_depend>ament_cmake_cpplint</exec_depend>
   <exec_depend>ament_cmake_flake8</exec_depend>
   <exec_depend>ament_cmake_pep257</exec_depend>
   <exec_depend>ament_cmake_uncrustify</exec_depend>
-  <!-- END -->  
+
   <exec_depend>ament_index_python</exec_depend>
   <exec_depend>python3-numpy</exec_depend>
   <exec_depend>rosidl_cli</exec_depend>
--- a/rosidl_generator_py/package.xml
+++ b/rosidl_generator_py/package.xml
@@ -29,6 +29,8 @@
   <buildtool_export_depend>rosidl_typesupport_c</buildtool_export_depend>
   <buildtool_export_depend>rosidl_typesupport_interface</buildtool_export_depend>
 
+  <exec_depend>ament_cmake_flake8</exec_depend>
+  <exec_depend>ament_cmake_pep257</exec_depend>
   <exec_depend>ament_index_python</exec_depend>
   <exec_depend>python3-numpy</exec_depend>
   <exec_depend>rosidl_cli</exec_depend>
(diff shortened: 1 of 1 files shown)


quarkytale:
Please add signature to your commit `git commit -s`

130s:
@quarkytale Thanks, done.

rosidl_generator_py/package.xml
```diff
@@ -29,6 +29,8 @@
   <buildtool_export_depend>rosidl_typesupport_c</buildtool_export_depend>
   <buildtool_export_depend>rosidl_typesupport_interface</buildtool_export_depend>
 
+  <exec_depend>ament_cmake_flake8</exec_depend>
+  <exec_depend>ament_cmake_pep257</exec_depend>
```
sloretz, line 33: Ah interesting. It seems like we would need a `test_export_depend` option to exist to describe these since they're needed only for tests in downstream packages. Adding them as `exec_depend` means that they'll always be installed when a message package is installed, but these dependencies are probably light enough that this is fine.

sloretz (approved):

rosidl_generator_py/package.xml
130s: > Adding them as `exec_depend` means that they'll always be installed when a message package is installed, but these dependencies are probably light enough that this is fine.

Ah, that makes sense (assuming you meant excluding test deps to be done by [`rosdep install -t`](https://github.com/ros-infrastructure/rosdep/pull/789)). Although I'm unsure how many users carefully exclude test deps, upstream pkg like this one should do the right thing. I'll update the PR.

> `test_export_depend`

:sweat_smile:

130s: Updated, please re-review.

rosidl_generator_py/package.xml
```diff
@@ -37,6 +37,8 @@
   <exec_depend>rosidl_runtime_c</exec_depend>
   <exec_depend>rpyutils</exec_depend>
 
+  <test_depend>ament_cmake_flake8</test_depend>
+  <test_depend>ament_cmake_pep257</test_depend>
```
sloretz, line 41: Ah, oops I wasn't requesting any changes. I think you were correct at first that these need to be `exec_depend` to make sure downstream message packages depend on them. 

I was only commenting to say the current package.xml depend declarations don't cover this case. We would need to create one, probably called `test_export_depend`, to support it better.

sloretz (changes requested):

rosidl_generator_py/package.xml
130s: Got it, reverted.

https://github.com/ros2/rosidl_python/pull/199#discussion_r1264225848
> Adding them as `exec_depend` means that they'll always be installed when a message package is installed, but these dependencies are probably light enough that this is fine.

It's a bit surprising no one (especially from those who are making products) has requested `test_export_depend`-like functionality. I can think of huge `test_depend`-ed packages in an upstream package.

clalancette (changes requested):

rosidl_generator_py/package.xml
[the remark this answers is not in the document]
130s: Those make sense, updated.

rosidl_generator_py/package.xml
```diff
@@ -29,6 +29,16 @@
...
+  <!-- BEGIN: Following 'exec_depend's are likely only needed in test in downstream,
+       so ideally "test_export_depend" (similar to 'build_export_depend') are reasonable.
+       With such a feature unexistent yet, they are 'exec_depend'-ed.
+       See https://github.com/ros2/rosidl_python/pull/199 -->
+  <exec_depend>ament_cmake_cppcheck</exec_depend>
+  <exec_depend>ament_cmake_cpplint</exec_depend>
+  <exec_depend>ament_cmake_flake8</exec_depend>
+  <exec_depend>ament_cmake_pep257</exec_depend>
+  <exec_depend>ament_cmake_uncrustify</exec_depend>
+  <!-- END -->
```
clalancette, line 41:
```suggestion
  <!-- TODO: The following 'exec_depend's are only needed in downstream tests,
       so ideally a 'test_export_depend' (similar to 'build_export_depend') would be used.
       Since that feature doesn't exist yet, they are 'exec_depend'-ed.
       See https://github.com/ros2/rosidl_python/pull/199 -->
  <exec_depend>ament_cmake_cppcheck</exec_depend>
  <exec_depend>ament_cmake_cpplint</exec_depend>
  <exec_depend>ament_cmake_flake8</exec_depend>
  <exec_depend>ament_cmake_pep257</exec_depend>
  <exec_depend>ament_cmake_uncrustify</exec_depend>

```

clalancette (changes requested):
Minor change to the comment here, then I'm happy with this.

PR #140, opened 2021-08-04
Prototype code for seeing if FindPython3 is usable for rosidl_python

sloretz:
This is a prototype intended to see how usable FindPython3 on non-linux platforms is. `rosidl_generator_py` makes the most use of `FindPythonExtra`.

Should inform https://github.com/ros2/python_cmake_module/pull/7

PR #201, opened 2024-01-17
Add include directories consistently PRIVATE instead of PUBLIC

cwecht:
This helps in situations in which:

1. If the numpy include directory is determined by executing python: https://github.com/ros2/rosidl_python/blob/65538f0f6bb74fcfa78c15427ad1fa2457bc15fd/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake#L174 **and**
2. If the path to your numpy installation may change.

This can happen e.g. if you want to cross-compile ROS2 **and** want to do it on some CICD setups.

The issue is that right now, e.g. for `std_msgs export_std_msgs__rosidl_generator_pyExport.cmake` sets the `INTERFACE_INCLUDE_DIRECTORIES `property of `std_msgs::std_msgs__rosidl_generator_py` to an absolute path (which causes issues if the folder of your numpy installation changes.

With this patch all include directories are consistently added `PRIVATE` instead of `PUBLIC`. There shouldn't be a need to expose them anyways.

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -189,7 +189,7 @@ if(APPLE OR WIN32 OR NOT _numpy_h)
       "error code ${_result}")
   endif()
   message(STATUS "Using numpy include directory: ${_output}")
-  target_include_directories(${_target_name_lib} PUBLIC "${_output}")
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
I'm going to ask to hold off on this one.  This change is also done by the much larger #140, but that one has additional benefits.  I'm just waiting for @sloretz to finish reviewing that one.

clalancette:
I've now merged in #140, which also contains this fix.  So I'll close this one out.

PR #206, opened 2024-03-15
Static typing for Message, Services, and Actions

### LINKED ISSUE #157: Python type annotation
#### Feature description
It would be useful if the python classes included type annotations. 

For example:

Parent.msg
```
Child[] children
Child child
```

Child.msg
```
uint8 some_child_content
```

Would generate something like:

```diff

+from perception_msgs.msg import Child

class Parent(metaclass=Metaclass_Parent):
    """Message class 'Parent'."""

    __slots__ = [
        '_children',
        '_child',
    ]

    _fields_and_field_types = {
        'children': 'sequence<perception_msgs/Child>',
        'child': 'perception_msgs/Child',
    }

    SLOT_TYPES = (
        rosidl_parser.definition.UnboundedSequence(rosidl_parser.definition.NamespacedType(['perception_msgs', 'msg'], 'Child')),  # noqa: E501
        rosidl_parser.definition.NamespacedType(['perception_msgs', 'msg'], 'Child'),  # noqa: E501
    )

-   def __init__(self, **kwargs):
+   def __init__(self, children: Optional[List[Child]]=None, child: Optional[Child]=None):
-       assert all('_' + key in self.__slots__ for key in kwargs.keys()), \
-           'Invalid arguments passed to constructor: %s' % \
-           ', '.join(sorted(k for k in kwargs.keys() if '_' + k not in self.__slots__))
-       self.children = kwargs.get('children', [])
-       from perception_msgs.msg import Child
-       self.child = kwargs.get('child', Child())
+       self.children = children or []
+       self.child = child or Child()

    def __repr__(self):
        typename = self.__class__.__module__.split('.')
        typename.pop()
        typename.append(self.__class__.__name__)
        args = []
        for s, t in zip(self.__slots__, self.SLOT_TYPES):
            field = getattr(self, s)
            fieldstr = repr(field)
            # We use Python array type for fields that can be directly stored
            # in them, and "normal" sequences for everything else.  If it is
            # a type that we store in an array, strip off the 'array' portion.
            if (
                isinstance(t, rosidl_parser.definition.AbstractSequence) and
                isinstance(t.value_type, rosidl_parser.definition.BasicType) and
                t.value_type.typename in ['float', 'double', 'int8', 'uint8', 'int16', 'uint16', 'int32', 'uint32', 'int64', 'uint64']
            ):
                if len(field) == 0:
                    fieldstr = '[]'
                else:
                    assert fieldstr.startswith('array(')
                    prefix = "array('X', "
                    suffix = ')'
                    fieldstr = fieldstr[len(prefix):-len(suffix)]
            args.append(s[1:] + '=' + fieldstr)
        return '%s(%s)' % ('.'.join(typename), ', '.join(args))

    def __eq__(self, other):
        if not isinstance(other, self.__class__):
            return False
        if self.children != other.children:
            return False
        if self.child != other.child:
            return False
        return True

    @classmethod
    def get_fields_and_field_types(cls):
        from copy import copy
        return copy(cls._fields_and_field_types)

    @property
-   def children(self):
+   def children(self) -> List[Child]:
        """Message field 'children'."""
        return self._children

    @children.setter
-   def children(self, value):
+   def children(self, value: List[Child]):
        if __debug__:
            from perception_msgs.msg import Child
            from collections.abc import Sequence
            from collections.abc import Set
            from collections import UserList
            from collections import UserString
            assert \
                ((isinstance(value, Sequence) or
                  isinstance(value, Set) or
                  isinstance(value, UserList)) and
                 not isinstance(value, str) and
                 not isinstance(value, UserString) and
                 all(isinstance(v, Child) for v in value) and
                 True), \
                "The 'children' field must be a set or sequence and each value of type 'Child'"
        self._children = value

    @property
-   def child(self):
+   def child(self) -> Child: 
        """Message field 'child'."""
        return self._child

    @child.setter
-   def child(self, value):
+   def child(self, value: Child):    
        if __debug__:
            from perception_msgs.msg import Child
            assert \
                isinstance(value, Child), \
                "The 'child' field must be a sub message of type 'Child'"
        self._child = value

```

#### Implementation considerations
* There seems to be an [issue](https://github.com/python/mypy/issues/220) with typing `@property` getters/setters. For me, using `pylance` the above syntax thinks `child` is of type `property` rather than type `Child`
* Backwards compatibility requirements - do we need to support python < 3.8? I assume type annotations break old python.
* Another path.... `from dataclasses import dataclass` gives us most of these features and more. Plus, it can be used with other libraries like [dacite](https://github.com/konradhalas/dacite)

InvincibleRMC:
Adds static typing to generated messages. Closes #157.


InvincibleRMC:
Still need to fix a circular import issue.

InvincibleRMC:
Running mypy complains about the type change within ` @property`  this doesn't effect the end user and is fine for now until [#mypy300](https://github.com/python/mypy/issues/3004) gets resolved.

haudren-woven:
Very nice! There is another working package here:
https://github.com/rospypi/rosidl_generator_mypy

I hope this can be merged and replace the above :pray: Thank you for working on type annotations.

InvincibleRMC:
> Very nice! There is another working package here: https://github.com/rospypi/rosidl_generator_mypy
> 
> I hope this can be merged and replace the above 🙏 Thank you for working on type annotations.

If your solutions works and gets merged in I will gladly close this pr.

haudren-woven:
Actually our solution is meant as a placeholder which you can use in your own workspace without modifying `rosidl_generator_py`. Last time we tried to PR this in, the maintainers did not seem to have much interest in merging it, but I'm supportive to merge your solution in: I'd rather focus our contributions on a core package such that it can benefit the larger community. From my cursory look, they look rather equivalent so let's go with it :+1:

InvincibleRMC:
One thing I have learn which is relevant to #216 is that currently according to pep [484](https://peps.python.org/pep-0484/#the-numeric-tower) is that a type annotation of `float` allows `int`. There is currently no defined way to define a `realFloat` type that disallows `int`. There are many discussions about [this](https://discuss.python.org/t/options-for-a-long-term-fix-of-the-special-case-for-float-int-complex/54289?page=2).

InvincibleRMC:
@sloretz @fujitatomoya could you review this pull request if you get the chance? It would improve the pub/sub, service/clients in `rclpy` since currently the generics are being ignored.

rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -21,7 +21,7 @@
 
 try:
     from .generate_py_impl import generate_py
-    assert generate_py
+    generate_py
```
sloretz: The line was `assert generate_py` before, but I'm not really sure that line was meaningful. I recommend deleting the line instead of leaving the statement `generate_py`.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -130,36 +130,42 @@ def print_warning_if_reserved_keyword(member_name, interface_type, interface_nam
                     sorted((value, key) for (key, value) in module_names.items()):
                 f.write(
                     f'from {package_name}.{subfolder}.{module_name} import '
-                    f'{idl_stem}  # noqa: F401\n')
+                    f'{idl_stem} as {idl_stem}  # noqa: F401\n')
```
sloretz, line 140: What does adding the `as` keyword here do? It looks like the name is unchanged in all of the lines in this file.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -192,6 +198,10 @@ def print_warning_if_reserved_keyword(member_name, interface_type, interface_nam
                 minimum_timestamp=latest_target_timestamp)
             generated_files.append(generated_file)
 
+    # Generate py.typed to mark the generate files as having type support as according to PEP561.
+    with open(os.path.join(args['output_dir'], 'py.typed'), 'w', encoding='utf-8'):
+        pass
```
sloretz: Recommend [`pathlib.Path.touch`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch) instead of `with open(...): pass`

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -91,22 +237,40 @@ for member in message.structure.members:
...
+        @(name.upper())__DEFAULT: @(type)
+@[     end for]@
+@[end if]@
+
     @@classmethod
-    def __import_type_support__(cls):
+    def __import_type_support__(cls) -> None:
         try:
-            from rosidl_generator_py import import_type_support
+            from rosidl_generator_py.rosidl_generator_py import import_type_support
```
sloretz: I think this import is incorrect

```python3
>>> from rosidl_generator_py.rosidl_generator_py import import_type_support
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'rosidl_generator_py.rosidl_generator_py'
>>> from rosidl_generator_py import import_type_support
>>> 
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -292,16 +453,33 @@ if isinstance(type_, AbstractNestedType):
...
+BUILTINS = [
+    a[0]
+    for a in inspect.getmembers(builtins)
+]
+}@
 
-    def __init__(self, **kwargs):
-        if 'check_fields' in kwargs:
-            self._check_fields = kwargs['check_fields']
+    def __init__(self,
```
sloretz: Please add `*,` here to force the arguments to be keyword only to avoid downstream code relying on the order of fields, and breaking if the message fields are added or reordered.

```suggestion
    def __init__(self, *,
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -312,8 +490,7 @@ if isinstance(type_, AbstractNestedType):
     type_ = type_.value_type
 }@
 @[  if member.has_annotation('default')]@
-        self.@(member.name) = kwargs.get(
-            '@(member.name)', @(message.structure.namespaced_type.name).@(member.name.upper())__DEFAULT)
+        self.@(member.name) = @(member.name) or @(message.structure.namespaced_type.name).@(member.name.upper())__DEFAULT
```
sloretz: Please check if the values are `None` instead of using `or`

```suggestion
        if 
        self.@(member.name) = @(member.name) if @(member.name) is not None else @(message.structure.namespaced_type.name).@(member.name.upper())__DEFAULT
```

Using `or` has an issue when default value evaluates to `True` but the value provided to `__init__` evaluates as False. For example, `Defaults.msg` has a field called `bool_value` with a default value of `true`: https://github.com/ros2/test_interface_files/blob/25d4b699e4bcd28ab4d900dd73540ea341b8c508/msg/Defaults.msg#L1

Using `or`, the line becomes:

```python3
self.bool_value = bool_value or Defaults.BOOL_VALUE__DEFAULT
```

If the user calls `Defaults(bool_value=False)`, then `self.bool_value = bool_value or Defaults.BOOL_VALUE__DEFAULT` becomes `self.bool_value = False or True` -> `self.bool_value = True`

sloretz (changes requested):
Thank you for the PR! It's no small feat to make changes to the empy templates 🙇 

I'm still looking through the changes in `_msg.py.em`, but I have some feedback in the meantime.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
InvincibleRMC: By default type checkers typically want imports from where classes are defined. So by default it would want `from rosidl_generator_py.msg._arrays import Arrays`. Using `as` re-exports it so it can be imported from a different file. This could also be exported using `__all__ = ['Arrays']`.

sloretz: Ah I did not know that!

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,46 +505,37 @@ if isinstance(type_, AbstractNestedType):
 @[    end if]@
 @[    if isinstance(member.type, Array)]@
 @[      if isinstance(type_, BasicType) and type_.typename == 'octet']@
-        self.@(member.name) = kwargs.get(
-            '@(member.name)',
-            [bytes([0]) for x in range(@(member.type.size))]
-        )
+        self.@(member.name) = @(member.name) or [bytes([0]) for x in range(@(member.type.size))]
```
sloretz:
```suggestion
        self.@(member.name) = @(member.name) if @(member.name) is not None else [bytes([0]) for x in range(@(member.type.size))]
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,46 +505,37 @@ if isinstance(type_, AbstractNestedType):
...
 @[      elif isinstance(type_, BasicType) and type_.typename in CHARACTER_TYPES]@
-        self.@(member.name) = kwargs.get(
-            '@(member.name)',
-            [chr(0) for x in range(@(member.type.size))]
-        )
+        self.@(member.name) = @(member.name) or [chr(0) for x in range(@(member.type.size))]
```
sloretz:
```suggestion
        self.@(member.name) = @(member.name) if @(member.name) is not None else [chr(0) for x in range(@(member.type.size))]
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,46 +505,37 @@ if isinstance(type_, AbstractNestedType):
...
         else:
-            self.@(member.name) = numpy.array(kwargs.get('@(member.name)'), dtype=@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['dtype']))
+            self.@(member.name) = numpy.array(@(member.name), dtype=@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['dtype']))
             assert self.@(member.name).shape == (@(member.type.size), )
 @[        else]@
-        self.@(member.name) = kwargs.get(
-            '@(member.name)',
-            [@(get_python_type(type_))() for x in range(@(member.type.size))]
-        )
+        self.@(member.name) = @(member.name) or [@(get_python_type(type_))() for x in range(@(member.type.size))]
```
sloretz:
```suggestion
        self.@(member.name) = @(member.name) if @(member.name) is not None else [@(get_python_type(type_))() for x in range(@(member.type.size))]
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,46 +505,37 @@ if isinstance(type_, AbstractNestedType):
...
 @[        end if]@
 @[      end if]@
 @[    elif isinstance(member.type, AbstractSequence)]@
 @[      if isinstance(member.type.value_type, BasicType) and member.type.value_type.typename in SPECIAL_NESTED_BASIC_TYPES]@
-        self.@(member.name) = array.array('@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['type_code'])', kwargs.get('@(member.name)', []))
+        self.@(member.name) = @(member.name) or array.array('@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['type_code'])', [])
```
sloretz:
```suggestion
        self.@(member.name) = @(member.name) if @(member.name) is not None else array.array('@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['type_code'])', [])
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,46 +505,37 @@ if isinstance(type_, AbstractNestedType):
...
 @[      else]@
-        self.@(member.name) = kwargs.get('@(member.name)', [])
+        self.@(member.name) = @(member.name) or []
```
sloretz:
```suggestion
        self.@(member.name) = @(member.name) if @(member.name) is not None else []
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,46 +505,37 @@ if isinstance(type_, AbstractNestedType):
...
 @[      end if]@
 @[    elif isinstance(type_, BasicType) and type_.typename == 'octet']@
-        self.@(member.name) = kwargs.get('@(member.name)', bytes([0]))
+        self.@(member.name) = @(member.name) or bytes([0])
```
sloretz:
```suggestion
        self.@(member.name) = @(member.name) if @(member.name) is not None else bytes([0])
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,46 +505,37 @@ if isinstance(type_, AbstractNestedType):
...
 @[    elif isinstance(type_, BasicType) and type_.typename in CHARACTER_TYPES]@
-        self.@(member.name) = kwargs.get('@(member.name)', chr(0))
+        self.@(member.name) = @(member.name) or chr(0)
```
sloretz:
```suggestion
        self.@(member.name) = @(member.name) if @(member.name) is not None else chr(0)
```

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,46 +505,37 @@ if isinstance(type_, AbstractNestedType):
...
 @[    else]@
-        self.@(member.name) = kwargs.get('@(member.name)', @(get_python_type(type_))())
+        self.@(member.name) = @(member.name) or @(get_python_type(type_))()
```
sloretz:
```suggestion
        self.@(member.name) = @(member.name) if @(member.name) is not None else @(get_python_type(type_))()
```

sloretz (commented):
Few more places to use `if @(member.name) is not None else` instead of `or`

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -29,6 +29,154 @@ from rosidl_parser.definition import SIGNED_INTEGER_TYPES
...
+}@
+if TYPE_CHECKING:
+    from ctypes import Structure
+
+    class PyCapsule(Structure):
+        pass  # don't need to define the full structure
+
+@[for type_import in type_imports]@
+    @(type_import)
+@[end for]
```
sloretz: Actions and services import the `_msg.py.em` template. That causes this block to get included in the file twice, and that causes linter test failures. You can reproduce by testing the `action_msgs` package with this PR.

Example: 
https://github.com/ros2/rosidl_python/blob/1d4a4c74c674558d781606be7e32cced4b6bc5f6/rosidl_generator_py/resource/_action.py.em#L8-L32

I'm not sure how to fix this. Maybe we could gather the type annotation information from all messages, and then generate the type annotation imports in a separate template? Something like

```python
type_annotation_info = gather_type_annotations(...)
TEMPLATE('_type_annotation_imports.py.em', 
 type_annotation_info=type_annotation_info, ...)
TEMPLATE('_msg_classes.py.em', 
 type_annotation_info=type_annotation_info, ...)
TEMPLATE('_msg_classes.py.em', 
 type_annotation_info=type_annotation_info, ...)
# ...
``` 

<details><summary>action_msgs.srv._cancel_goal.py</summary>

```python
# generated from rosidl_generator_py/resource/_idl.py.em
# with input from action_msgs:srv/CancelGoal.idl
# generated code does not contain a copyright notice

# This is being done at the module level and not on the instance level to avoid looking
# for the same variable multiple times on each instance. This variable is not supposed to
# change during runtime so it makes sense to only look for it once.
from collections.abc import MutableMapping
from os import getenv
from typing import Any, ClassVar, Dict, Optional, Tuple, Type, TYPE_CHECKING, TypedDict

ros_python_check_fields = getenv('ROS_PYTHON_CHECK_FIELDS', default='')
if TYPE_CHECKING:
    from ctypes import Structure

    class PyCapsule(Structure):
        pass  # don't need to define the full structure

    from action_msgs.msg import GoalInfo

# Import statements for member types

import builtins  # noqa: E402, I100

import rosidl_parser.definition  # noqa: E402, I100


class Metaclass_CancelGoal_Request(type):
    """Metaclass of message 'CancelGoal_Request'."""

    _CREATE_ROS_MESSAGE: ClassVar[Optional['PyCapsule']] = None
    _CONVERT_FROM_PY: ClassVar[Optional['PyCapsule']] = None
    _CONVERT_TO_PY: ClassVar[Optional['PyCapsule']] = None
    _DESTROY_ROS_MESSAGE: ClassVar[Optional['PyCapsule']] = None
    _TYPE_SUPPORT: ClassVar[Optional['PyCapsule']] = None

    class CancelGoal_RequestConstants(TypedDict):
        pass

    __constants: CancelGoal_RequestConstants = {
    }

    class CancelGoal_RequestDefault(CancelGoal_RequestConstants):
        pass

    @classmethod
    def __import_type_support__(cls) -> None:
        try:
            from rosidl_generator_py import import_type_support
            module = import_type_support('action_msgs')
        except ImportError:
            import logging
            import traceback
            logger = logging.getLogger(
                'action_msgs.srv.CancelGoal_Request')
            logger.debug(
                'Failed to import needed modules for type support:\n' +
                traceback.format_exc())
        else:
            cls._CREATE_ROS_MESSAGE = module.create_ros_message_msg__srv__cancel_goal__request
            cls._CONVERT_FROM_PY = module.convert_from_py_msg__srv__cancel_goal__request
            cls._CONVERT_TO_PY = module.convert_to_py_msg__srv__cancel_goal__request
            cls._TYPE_SUPPORT = module.type_support_msg__srv__cancel_goal__request
            cls._DESTROY_ROS_MESSAGE = module.destroy_ros_message_msg__srv__cancel_goal__request

            from action_msgs.msg import GoalInfo
            if GoalInfo._TYPE_SUPPORT is None:
                GoalInfo.__import_type_support__()

    @classmethod
    def __prepare__(metacls, name: str, bases: Tuple[Type[Any], ...], /, **kwds: Any) -> MutableMapping[str, object]:
        # list constant names here so that they appear in the help text of
        # the message class under "Data and other attributes defined here:"
        # as well as populate each message instance
        return {
        }


class CancelGoal_Request(metaclass=Metaclass_CancelGoal_Request):
    """Message class 'CancelGoal_Request'."""

    __slots__ = [
        '_goal_info',
        '_check_fields',
    ]

    _fields_and_field_types: Dict[str, str] = {
        'goal_info': 'action_msgs/GoalInfo',
    }

    # This attribute is used to store an rosidl_parser.definition variable
    # related to the data type of each of the components the message.
    SLOT_TYPES: Tuple[rosidl_parser.definition.AbstractType, ...] = (
        rosidl_parser.definition.NamespacedType(['action_msgs', 'msg'], 'GoalInfo'),  # noqa: E501
    )

    def __init__(self, *,
                 goal_info: Optional['GoalInfo'] = None,  # noqa: E501
                 check_fields: Optional[bool] = None) -> None:
        if check_fields is not None:
            self._check_fields = check_fields
        else:
            self._check_fields = ros_python_check_fields == '1'
        from action_msgs.msg import GoalInfo
        self.goal_info = goal_info if goal_info is not None else GoalInfo()

    def __repr__(self) -> str:
        typename = self.__class__.__module__.split('.')
        typename.pop()
        args = []
        for s, t in zip(self.get_fields_and_field_types().keys(), self.SLOT_TYPES):
            field = getattr(self, s)
            fieldstr = repr(field)
            # We use Python array type for fields that can be directly stored
            # in them, and "normal" sequences for everything else.  If it is
            # a type that we store in an array, strip off the 'array' portion.
            if (
                isinstance(t, rosidl_parser.definition.AbstractSequence) and
                isinstance(t.value_type, rosidl_parser.definition.BasicType) and
                t.value_type.typename in ['float', 'double', 'int8', 'uint8', 'int16', 'uint16', 'int32', 'uint32', 'int64', 'uint64']
            ):
                if len(field) == 0:
                    fieldstr = '[]'
                else:
                    if self._check_fields:
                        assert fieldstr.startswith('array(')
                    prefix = "array('X', "
                    suffix = ')'
                    fieldstr = fieldstr[len(prefix):-len(suffix)]
            args.append(s + '=' + fieldstr)
        return '%s(%s)' % ('.'.join(typename), ', '.join(args))

    def __eq__(self, other: object) -> bool:
        if not isinstance(other, CancelGoal_Request):
            return False
        if self.goal_info != other.goal_info:
            return False
        return True

    @classmethod
    def get_fields_and_field_types(cls) -> Dict[str, str]:
        from copy import copy
        return copy(cls._fields_and_field_types)

    @builtins.property
    def goal_info(self) -> 'GoalInfo':
        """Message field 'goal_info'."""
        return self._goal_info

    @goal_info.setter
    def goal_info(self, value: 'GoalInfo') -> None:
        if self._check_fields:
            from action_msgs.msg import GoalInfo
            assert \
                isinstance(value, GoalInfo), \
                "The 'goal_info' field must be a sub message of type 'GoalInfo'"
        self._goal_info = value
if TYPE_CHECKING:
    from ctypes import Structure

    class PyCapsule(Structure):
        pass  # don't need to define the full structure

    from action_msgs.msg import GoalInfo
    from typing import Literal
    from collections.abc import Sequence
    from typing import Union

# Import statements for member types

# already imported above
# import builtins

# already imported above
# import rosidl_parser.definition


class Metaclass_CancelGoal_Response(type):
    """Metaclass of message 'CancelGoal_Response'."""

    _CREATE_ROS_MESSAGE: ClassVar[Optional['PyCapsule']] = None
    _CONVERT_FROM_PY: ClassVar[Optional['PyCapsule']] = None
    _CONVERT_TO_PY: ClassVar[Optional['PyCapsule']] = None
    _DESTROY_ROS_MESSAGE: ClassVar[Optional['PyCapsule']] = None
    _TYPE_SUPPORT: ClassVar[Optional['PyCapsule']] = None

    class CancelGoal_ResponseConstants(TypedDict):
        ERROR_NONE: 'Literal[0]'
        ERROR_REJECTED: 'Literal[1]'
        ERROR_UNKNOWN_GOAL_ID: 'Literal[2]'
        ERROR_GOAL_TERMINATED: 'Literal[3]'

    __constants: CancelGoal_ResponseConstants = {
        'ERROR_NONE': 0,
        'ERROR_REJECTED': 1,
        'ERROR_UNKNOWN_GOAL_ID': 2,
        'ERROR_GOAL_TERMINATED': 3,
    }

    class CancelGoal_ResponseDefault(CancelGoal_ResponseConstants):
        pass

    @classmethod
    def __import_type_support__(cls) -> None:
        try:
            from rosidl_generator_py import import_type_support
            module = import_type_support('action_msgs')
        except ImportError:
            import logging
            import traceback
            logger = logging.getLogger(
                'action_msgs.srv.CancelGoal_Response')
            logger.debug(
                'Failed to import needed modules for type support:\n' +
                traceback.format_exc())
        else:
            cls._CREATE_ROS_MESSAGE = module.create_ros_message_msg__srv__cancel_goal__response
            cls._CONVERT_FROM_PY = module.convert_from_py_msg__srv__cancel_goal__response
            cls._CONVERT_TO_PY = module.convert_to_py_msg__srv__cancel_goal__response
            cls._TYPE_SUPPORT = module.type_support_msg__srv__cancel_goal__response
            cls._DESTROY_ROS_MESSAGE = module.destroy_ros_message_msg__srv__cancel_goal__response

            from action_msgs.msg import GoalInfo
            if GoalInfo._TYPE_SUPPORT is None:
                GoalInfo.__import_type_support__()

    @classmethod
    def __prepare__(metacls, name: str, bases: Tuple[Type[Any], ...], /, **kwds: Any) -> MutableMapping[str, object]:
        # list constant names here so that they appear in the help text of
        # the message class under "Data and other attributes defined here:"
        # as well as populate each message instance
        return {
            'ERROR_NONE': metacls.__constants['ERROR_NONE'],
            'ERROR_REJECTED': metacls.__constants['ERROR_REJECTED'],
            'ERROR_UNKNOWN_GOAL_ID': metacls.__constants['ERROR_UNKNOWN_GOAL_ID'],
            'ERROR_GOAL_TERMINATED': metacls.__constants['ERROR_GOAL_TERMINATED'],
        }

    @property
    def ERROR_NONE(self) -> 'Literal[0]':
        """Message constant 'ERROR_NONE'."""
        return Metaclass_CancelGoal_Response.__constants['ERROR_NONE']

    @property
    def ERROR_REJECTED(self) -> 'Literal[1]':
        """Message constant 'ERROR_REJECTED'."""
        return Metaclass_CancelGoal_Response.__constants['ERROR_REJECTED']

    @property
    def ERROR_UNKNOWN_GOAL_ID(self) -> 'Literal[2]':
        """Message constant 'ERROR_UNKNOWN_GOAL_ID'."""
        return Metaclass_CancelGoal_Response.__constants['ERROR_UNKNOWN_GOAL_ID']

    @property
    def ERROR_GOAL_TERMINATED(self) -> 'Literal[3]':
        """Message constant 'ERROR_GOAL_TERMINATED'."""
        return Metaclass_CancelGoal_Response.__constants['ERROR_GOAL_TERMINATED']


class CancelGoal_Response(metaclass=Metaclass_CancelGoal_Response):
    """
    Message class 'CancelGoal_Response'.

    Constants:
      ERROR_NONE
      ERROR_REJECTED
      ERROR_UNKNOWN_GOAL_ID
      ERROR_GOAL_TERMINATED
    """

    __slots__ = [
        '_return_code',
        '_goals_canceling',
        '_check_fields',
    ]

    _fields_and_field_types: Dict[str, str] = {
        'return_code': 'int8',
        'goals_canceling': 'sequence<action_msgs/GoalInfo>',
    }

    # This attribute is used to store an rosidl_parser.definition variable
    # related to the data type of each of the components the message.
    SLOT_TYPES: Tuple[rosidl_parser.definition.AbstractType, ...] = (
        rosidl_parser.definition.BasicType('int8'),  # noqa: E501
        rosidl_parser.definition.UnboundedSequence(rosidl_parser.definition.NamespacedType(['action_msgs', 'msg'], 'GoalInfo')),  # noqa: E501
    )

    def __init__(self, *,
                 return_code: Optional['int'] = None,  # noqa: E501
                 goals_canceling: Optional['Union[Sequence[GoalInfo], Set[GoalInfo], UserList[GoalInfo]]'] = None,  # noqa: E501
                 check_fields: Optional[bool] = None) -> None:
        if check_fields is not None:
            self._check_fields = check_fields
        else:
            self._check_fields = ros_python_check_fields == '1'
        self.return_code = return_code if return_code is not None else int()
        self.goals_canceling = goals_canceling if goals_canceling is not None else []

    def __repr__(self) -> str:
        typename = self.__class__.__module__.split('.')
        typename.pop()
        args = []
        for s, t in zip(self.get_fields_and_field_types().keys(), self.SLOT_TYPES):
            field = getattr(self, s)
            fieldstr = repr(field)
            # We use Python array type for fields that can be directly stored
            # in them, and "normal" sequences for everything else.  If it is
            # a type that we store in an array, strip off the 'array' portion.
            if (
                isinstance(t, rosidl_parser.definition.AbstractSequence) and
                isinstance(t.value_type, rosidl_parser.definition.BasicType) and
                t.value_type.typename in ['float', 'double', 'int8', 'uint8', 'int16', 'uint16', 'int32', 'uint32', 'int64', 'uint64']
            ):
                if len(field) == 0:
                    fieldstr = '[]'
                else:
                    if self._check_fields:
                        assert fieldstr.startswith('array(')
                    prefix = "array('X', "
                    suffix = ')'
                    fieldstr = fieldstr[len(prefix):-len(suffix)]
            args.append(s + '=' + fieldstr)
        return '%s(%s)' % ('.'.join(typename), ', '.join(args))

    def __eq__(self, other: object) -> bool:
        if not isinstance(other, CancelGoal_Response):
            return False
        if self.return_code != other.return_code:
            return False
        if self.goals_canceling != other.goals_canceling:
            return False
        return True

    @classmethod
    def get_fields_and_field_types(cls) -> Dict[str, str]:
        from copy import copy
        return copy(cls._fields_and_field_types)

    @builtins.property
    def return_code(self) -> 'int':
        """Message field 'return_code'."""
        return self._return_code

    @return_code.setter
    def return_code(self, value: 'int') -> None:
        if self._check_fields:
            assert \
                isinstance(value, int), \
                "The 'return_code' field must be of type 'int'"
            assert value >= -128 and value < 128, \
                "The 'return_code' field must be an integer in [-128, 127]"
        self._return_code = value

    @builtins.property
    def goals_canceling(self) -> 'Union[Sequence[GoalInfo], Set[GoalInfo], UserList[GoalInfo]]':
        """Message field 'goals_canceling'."""
        return self._goals_canceling

    @goals_canceling.setter
    def goals_canceling(self, value: 'Union[Sequence[GoalInfo], Set[GoalInfo], UserList[GoalInfo]]') -> None:
        if self._check_fields:
            from action_msgs.msg import GoalInfo
            from collections.abc import Sequence
            assert \
                ((isinstance(value, Sequence) or
                  isinstance(value, Set) or
                  isinstance(value, UserList)) and
                 not isinstance(value, str) and
                 not isinstance(value, UserString) and
                 all(isinstance(v, GoalInfo) for v in value) and
                 True), \
                "The 'goals_canceling' field must be a set or sequence and each value of type 'GoalInfo'"
        self._goals_canceling = value
if TYPE_CHECKING:
    from ctypes import Structure

    class PyCapsule(Structure):
        pass  # don't need to define the full structure

    from action_msgs.srv import CancelGoal_Response
    from service_msgs.msg import ServiceEventInfo
    from action_msgs.srv import CancelGoal_Request
    from collections.abc import Sequence
    from typing import Union

# Import statements for member types

# already imported above
# import builtins

# already imported above
# import rosidl_parser.definition


class Metaclass_CancelGoal_Event(type):
    """Metaclass of message 'CancelGoal_Event'."""

    _CREATE_ROS_MESSAGE: ClassVar[Optional['PyCapsule']] = None
    _CONVERT_FROM_PY: ClassVar[Optional['PyCapsule']] = None
    _CONVERT_TO_PY: ClassVar[Optional['PyCapsule']] = None
    _DESTROY_ROS_MESSAGE: ClassVar[Optional['PyCapsule']] = None
    _TYPE_SUPPORT: ClassVar[Optional['PyCapsule']] = None

    class CancelGoal_EventConstants(TypedDict):
        pass

    __constants: CancelGoal_EventConstants = {
    }

    class CancelGoal_EventDefault(CancelGoal_EventConstants):
        pass

    @classmethod
    def __import_type_support__(cls) -> None:
        try:
            from rosidl_generator_py import import_type_support
            module = import_type_support('action_msgs')
        except ImportError:
            import logging
            import traceback
            logger = logging.getLogger(
                'action_msgs.srv.CancelGoal_Event')
            logger.debug(
                'Failed to import needed modules for type support:\n' +
                traceback.format_exc())
        else:
            cls._CREATE_ROS_MESSAGE = module.create_ros_message_msg__srv__cancel_goal__event
            cls._CONVERT_FROM_PY = module.convert_from_py_msg__srv__cancel_goal__event
            cls._CONVERT_TO_PY = module.convert_to_py_msg__srv__cancel_goal__event
            cls._TYPE_SUPPORT = module.type_support_msg__srv__cancel_goal__event
            cls._DESTROY_ROS_MESSAGE = module.destroy_ros_message_msg__srv__cancel_goal__event

            from service_msgs.msg import ServiceEventInfo
            if ServiceEventInfo._TYPE_SUPPORT is None:
                ServiceEventInfo.__import_type_support__()

    @classmethod
    def __prepare__(metacls, name: str, bases: Tuple[Type[Any], ...], /, **kwds: Any) -> MutableMapping[str, object]:
        # list constant names here so that they appear in the help text of
        # the message class under "Data and other attributes defined here:"
        # as well as populate each message instance
        return {
        }


class CancelGoal_Event(metaclass=Metaclass_CancelGoal_Event):
    """Message class 'CancelGoal_Event'."""

    __slots__ = [
        '_info',
        '_request',
        '_response',
        '_check_fields',
    ]

    _fields_and_field_types: Dict[str, str] = {
        'info': 'service_msgs/ServiceEventInfo',
        'request': 'sequence<action_msgs/CancelGoal_Request, 1>',
        'response': 'sequence<action_msgs/CancelGoal_Response, 1>',
    }

    # This attribute is used to store an rosidl_parser.definition variable
    # related to the data type of each of the components the message.
    SLOT_TYPES: Tuple[rosidl_parser.definition.AbstractType, ...] = (
        rosidl_parser.definition.NamespacedType(['service_msgs', 'msg'], 'ServiceEventInfo'),  # noqa: E501
        rosidl_parser.definition.BoundedSequence(rosidl_parser.definition.NamespacedType(['action_msgs', 'srv'], 'CancelGoal_Request'), 1),  # noqa: E501
        rosidl_parser.definition.BoundedSequence(rosidl_parser.definition.NamespacedType(['action_msgs', 'srv'], 'CancelGoal_Response'), 1),  # noqa: E501
    )

    def __init__(self, *,
                 info: Optional['ServiceEventInfo'] = None,  # noqa: E501
                 request: Optional['Union[Sequence[CancelGoal_Request], Set[CancelGoal_Request], UserList[CancelGoal_Request]]'] = None,  # noqa: E501
                 response: Optional['Union[Sequence[CancelGoal_Response], Set[CancelGoal_Response], UserList[CancelGoal_Response]]'] = None,  # noqa: E501
                 check_fields: Optional[bool] = None) -> None:
        if check_fields is not None:
            self._check_fields = check_fields
        else:
            self._check_fields = ros_python_check_fields == '1'
        from service_msgs.msg import ServiceEventInfo
        self.info = info if info is not None else ServiceEventInfo()
        self.request = request if request is not None else []
        self.response = response if response is not None else []

    def __repr__(self) -> str:
        typename = self.__class__.__module__.split('.')
        typename.pop()
        args = []
        for s, t in zip(self.get_fields_and_field_types().keys(), self.SLOT_TYPES):
            field = getattr(self, s)
            fieldstr = repr(field)
            # We use Python array type for fields that can be directly stored
            # in them, and "normal" sequences for everything else.  If it is
            # a type that we store in an array, strip off the 'array' portion.
            if (
                isinstance(t, rosidl_parser.definition.AbstractSequence) and
                isinstance(t.value_type, rosidl_parser.definition.BasicType) and
                t.value_type.typename in ['float', 'double', 'int8', 'uint8', 'int16', 'uint16', 'int32', 'uint32', 'int64', 'uint64']
            ):
                if len(field) == 0:
                    fieldstr = '[]'
                else:
                    if self._check_fields:
                        assert fieldstr.startswith('array(')
                    prefix = "array('X', "
                    suffix = ')'
                    fieldstr = fieldstr[len(prefix):-len(suffix)]
            args.append(s + '=' + fieldstr)
        return '%s(%s)' % ('.'.join(typename), ', '.join(args))

    def __eq__(self, other: object) -> bool:
        if not isinstance(other, CancelGoal_Event):
            return False
        if self.info != other.info:
            return False
        if self.request != other.request:
            return False
        if self.response != other.response:
            return False
        return True

    @classmethod
    def get_fields_and_field_types(cls) -> Dict[str, str]:
        from copy import copy
        return copy(cls._fields_and_field_types)

    @builtins.property
    def info(self) -> 'ServiceEventInfo':
        """Message field 'info'."""
        return self._info

    @info.setter
    def info(self, value: 'ServiceEventInfo') -> None:
        if self._check_fields:
            from service_msgs.msg import ServiceEventInfo
            assert \
                isinstance(value, ServiceEventInfo), \
                "The 'info' field must be a sub message of type 'ServiceEventInfo'"
        self._info = value

    @builtins.property
    def request(self) -> 'Union[Sequence[CancelGoal_Request], Set[CancelGoal_Request], UserList[CancelGoal_Request]]':
        """Message field 'request'."""
        return self._request

    @request.setter
    def request(self, value: 'Union[Sequence[CancelGoal_Request], Set[CancelGoal_Request], UserList[CancelGoal_Request]]') -> None:
        if self._check_fields:
            from action_msgs.srv import CancelGoal_Request
            from collections.abc import Sequence
            assert \
                ((isinstance(value, Sequence) or
                  isinstance(value, Set) or
                  isinstance(value, UserList)) and
                 not isinstance(value, str) and
                 not isinstance(value, UserString) and
                 len(value) <= 1 and
                 all(isinstance(v, CancelGoal_Request) for v in value) and
                 True), \
                "The 'request' field must be a set or sequence with length <= 1 and each value of type 'CancelGoal_Request'"
        self._request = value

    @builtins.property
    def response(self) -> 'Union[Sequence[CancelGoal_Response], Set[CancelGoal_Response], UserList[CancelGoal_Response]]':
        """Message field 'response'."""
        return self._response

    @response.setter
    def response(self, value: 'Union[Sequence[CancelGoal_Response], Set[CancelGoal_Response], UserList[CancelGoal_Response]]') -> None:
        if self._check_fields:
            from action_msgs.srv import CancelGoal_Response
            from collections.abc import Sequence
            assert \
                ((isinstance(value, Sequence) or
                  isinstance(value, Set) or
                  isinstance(value, UserList)) and
                 not isinstance(value, str) and
                 not isinstance(value, UserString) and
                 len(value) <= 1 and
                 all(isinstance(v, CancelGoal_Response) for v in value) and
                 True), \
                "The 'response' field must be a set or sequence with length <= 1 and each value of type 'CancelGoal_Response'"
        self._response = value

class Metaclass_CancelGoal(type):
    """Metaclass of service 'CancelGoal'."""

    _TYPE_SUPPORT = None

    @classmethod
    def __import_type_support__(cls) -> None:
        try:
            from rosidl_generator_py import import_type_support
            module = import_type_support('action_msgs')
        except ImportError:
            import logging
            import traceback
            logger = logging.getLogger(
                'action_msgs.srv.CancelGoal')
            logger.debug(
                'Failed to import needed modules for type support:\n' +
                traceback.format_exc())
        else:
            cls._TYPE_SUPPORT = module.type_support_srv__srv__cancel_goal

            from action_msgs.srv import _cancel_goal
            if _cancel_goal.Metaclass_CancelGoal_Request._TYPE_SUPPORT is None:
                _cancel_goal.Metaclass_CancelGoal_Request.__import_type_support__()
            if _cancel_goal.Metaclass_CancelGoal_Response._TYPE_SUPPORT is None:
                _cancel_goal.Metaclass_CancelGoal_Response.__import_type_support__()
            if _cancel_goal.Metaclass_CancelGoal_Event._TYPE_SUPPORT is None:
                _cancel_goal.Metaclass_CancelGoal_Event.__import_type_support__()


class CancelGoal(metaclass=Metaclass_CancelGoal):
    from action_msgs.srv._cancel_goal import CancelGoal_Request as Request
    from action_msgs.srv._cancel_goal import CancelGoal_Response as Response
    from action_msgs.srv._cancel_goal import CancelGoal_Event as Event

    def __init__(self) -> None:
        raise NotImplementedError('Service classes can not be instantiated')

```
</details>

sloretz: I guess there's already a pattern to fix this, it's the passing in of import statements into the template:

https://github.com/ros2/rosidl_python/blob/1d4a4c74c674558d781606be7e32cced4b6bc5f6/rosidl_generator_py/resource/_msg.py.em#L77-L78

If we keep a `type_annotations_import_statements` list, then we can do the same thing.

InvincibleRMC: I have started working on this. Will push once it works.

InvincibleRMC:
@sloretz Following up I was able to remove the duplicate imports. Let me know if there are any other questions/suggestions. Thanks again for reviewing this pr.

InvincibleRMC:
Now that kilted is starting to wrap up friendly ping @sloretz to re-review when you get a chance.

InvincibleRMC:
Friendly ping @sloretz to finish the review when you get a chance.

ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.ros.org/t/python-generated-messages-with-static-type-hints/45156/1

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -321,3 +336,107 @@ def get_python_type(type_):
...
+    elif isinstance(member.type, AbstractGenericString) and member.type.has_maximum_size():
+        type_annotation = 'typing.Union[str, collections.UserString]'
+
+        type_imports.add('import collections')
+    elif isinstance(type_, BasicType) and type_.typename == 'char':
+        type_annotation = 'typing.Union[str, collection.UserString]'
+
+        type_imports.add('import collections')
+    elif isinstance(type_, BasicType) and type_.typename == 'octet':
+        type_annotation = 'typing.Union[bytes, collections.abc.ByteString]'
```
benjaminwp18: Should the duplicate `isinstance(member.type, AbstractNestedType)` and `isinstance(type_, BasicType)` cases be combined here?

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -321,3 +336,107 @@ def get_python_type(type_):
...
+
+    if isinstance(constant.type, AbstractNestedType) and isinstance(type_, BasicType) and \
+       type_.typename in SPECIAL_NESTED_BASIC_TYPES:
+        if isinstance(constant.type, Array):
+            type_imports.add('import numpy.typing')
+            dtype = SPECIAL_NESTED_BASIC_TYPES[type_.typename]['dtype']
+            return f'numpy.typing.NDArray[{dtype}]'
+        elif isinstance(constant.type, AbstractSequence):
+            return f'array.array[{python_type}]'
+    elif isinstance(constant.type, AbstractNestedType):
```
benjaminwp18: Should the duplicate `isinstance(constant.type, AbstractNestedType)` cases be combined here?

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -321,3 +336,107 @@ def get_python_type(type_):
         return 'str'
 
     assert False, "unknown type '%s'" % type_
+
+
+def get_type_annotation_constant_default(constant: Constant, value: Union[str, int, float, bool],
```
benjaminwp18: Consider adding a docstring

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -321,3 +336,107 @@ def get_python_type(type_):
...
+            return 'float'
+        elif isinstance(type_, BasicType) and type_.typename == 'octet':
+            return 'bytes'
+        else:
+            return f'typing.Literal[{value}]'
+
+    assert False, f"unknown type '{type_}'"
+
+
+def get_setter_and_getter_type(member: Member, type_imports: set[str]) -> tuple[str, str]:
```
benjaminwp18, line 390: Consider adding a docstring

benjaminwp18 (changes requested):

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
InvincibleRMC: Since isinstance checks do type narrowing combining them would lose some typing information.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
InvincibleRMC: Since isinstance checks do type narrowing combining them would lose some typing information.

InvincibleRMC:
@benjaminwp18 Thanks so much for the review! I have addressed your feedback. Please let me know if you have any other feedback.

InvincibleRMC:
@christophebedard when you get a chance I would appreciate a ROS Maintainer review as well.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -292,16 +346,33 @@ if isinstance(type_, AbstractNestedType):
...
+@{
+# Taken from flake8-builtins
+# https://github.com/gforcada/flake8-builtins/blob/689ad01c03cb52ae73be23d19706e6a4a491f4e9/flake8_builtins.py
+import inspect
+import builtins
+BUILTINS = [
+    a[0]
+    for a in inspect.getmembers(builtins)
+]
+}@
```
christophebedard: This is pretty dangerous from a legal/licensing point of view, since the `flake8-builtins` code is [GPLv2-licensed](https://github.com/gforcada/flake8-builtins/blob/689ad01c03cb52ae73be23d19706e6a4a491f4e9/LICENSE). GPLv2 and other copyleft licenses require that any "derivative work" be also licensed GPLv2 ("GPL virality").

Please rewrite this so that the code isn't just copied from a GPLv2-licensed file.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -66,8 +116,6 @@ for member in message.structure.members:
 @
 @#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 @[if imports]@
-
-
```
christophebedard: What is the reason for removing this? I think it's meant to keep a normal 2-line separation between sections in the generated file(?).

InvincibleRMC: added back

rosidl_generator_py/resource/_msg.py.em
InvincibleRMC: Ok I rewrote it based on the already existing builtins handler in `_msg.py.em` seen [here](https://github.com/ros2/rosidl_python/blob/4dda0a4777dff29d1e95d6123cc38c6d9a9bdba3/rosidl_generator_py/resource/_msg.py.em#L423-L432)

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -301,5 +301,10 @@ if(BUILD_TESTING AND rosidl_generate_interfaces_ADD_LINTER_TESTS)
       # a value of zero tells uncrustify to ignore line length
       MAX_LINE_LENGTH 0
       "${_output_path}")
+
+    find_package(ament_cmake_mypy REQUIRED)
+    ament_mypy(
+      TESTNAME "mypy_rosidl_generate_py"
```
christophebedard: To match the other `*_rosidl_generated_py` test names:

```suggestion
      TESTNAME "mypy_rosidl_generated_py"
```

rosidl_generator_py/resource/_idl.py.em
```diff
@@ -5,7 +5,12 @@
...
 # for the same variable multiple times on each instance. This variable is not supposed to
 # change during runtime so it makes sense to only look for it once.
+
+from __future__ import annotations
+
+import collections.abc
 from os import getenv
+import typing
 
 ros_python_check_fields = getenv('ROS_PYTHON_CHECK_FIELDS', default='')
```
christophebedard: I think the comment is referring to this `ros_python_check_fields` line. Since you're adding some imports, can you move the 3 comment lines so that they are above this line? Otherwise it might be a bit confusing

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -321,3 +337,115 @@ def get_python_type(type_):
...
+    """From a constant/member return type annotations for constant/default values."""
+    type_ = constant.type
+
+    if isinstance(type_, AbstractNestedType):
+        type_ = type_.value_type
+
+    python_type = get_python_type(type_)
+
+    if isinstance(constant.type, AbstractNestedType) and isinstance(type_, BasicType) and \
+       type_.typename in SPECIAL_NESTED_BASIC_TYPES:
```
christophebedard: Minor comment, feel free to ignore:

Personally, I'm a fan of `if (...):` for readability (I think the `and` goes at the end of the previous line and not at the beginning of the next line, but I'm not 100% sure):

```suggestion
    if (
        isinstance(constant.type, AbstractNestedType) and isinstance(type_, BasicType) and
        type_.typename in SPECIAL_NESTED_BASIC_TYPES
    ):
```

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -321,3 +337,115 @@ def get_python_type(type_):
...
+    if isinstance(type_, AbstractNestedType):
+        type_ = type_.value_type
+
+    python_type = get_python_type(type_)
+
+    type_annotation = ''
+    type_annotations_getter = ''
+
+    if isinstance(member.type, AbstractNestedType) and isinstance(type_, BasicType) and \
+       type_.typename in SPECIAL_NESTED_BASIC_TYPES:
```
christophebedard: Similar to my other minor comment, I'd suggest using `if (...):`, but it's up to you.

rosidl_generator_py/resource/_action.py.em
```diff
@@ -5,40 +5,50 @@ from rosidl_pycommon import convert_camel_case_to_lower_case_underscore
 action_name = '_' + convert_camel_case_to_lower_case_underscore(action.namespaced_type.name)
 module_name = '_' + convert_camel_case_to_lower_case_underscore(interface_path.stem)
 
+type_annotations_import_statements.add(f'from {'.'.join(action.namespaced_type.namespaces)}.{module_name} import {action.goal.structure.namespaced_type.name}')
+type_annotations_import_statements.add(f'from {'.'.join(action.namespaced_type.namespaces)}.{module_name} import {action.result.structure.namespaced_type.name}')
+type_annotations_import_statements.add(f'from {'.'.join(action.namespaced_type.namespaces)}.{module_name} import {action.feedback.structure.namespaced_type.name}')
```
christophebedard: Don't you need to use double quotes inside of the single-quote f-string?

InvincibleRMC: From what I can tell its fine. Even on older python versions. https://mypy-play.net/?mypy=1.17.0&python=3.9&gist=701dd3aa4272078921dca20a46e1eb04

christophebedard: :exploding_head: alright then!

christophebedard:
Pulls: ros2/rosidl_python#206
Gist: https://gist.githubusercontent.com/christophebedard/9a392164fa167b98f745f9a505a40a73/raw/1cb2fe9a7ed734bbd4a459211613e40209a97991/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16562
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24404/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18474/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3849/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24969/)

christophebedard:
I ran full CI^ so that interfaces get generated, and so that hopefully the linters (including mypy) run on the generated files as part of tests in the various interface packages. The Rpr job doesn't really test much here.

InvincibleRMC:
Thanks so much for the review!

> I ran full CI^ so that interfaces get generated, and so that hopefully the linters (including mypy) run on the generated files as part of tests in the various interface packages. The Rpr job doesn't really test much here.

I was going to ask about this since I have a feeling this will expose some downstream bugs to be fixed.

rosidl_generator_py/resource/_action.py.em
InvincibleRMC: Apparently I'm incorrect and also seem to misunderstand how mypy playground works. Will fix `rhel`

christophebedard:
Pulls: ros2/rosidl_python#206
Gist: https://gist.githubusercontent.com/christophebedard/23340b5f0c0b1e577396c407ded46165/raw/1cb2fe9a7ed734bbd4a459211613e40209a97991/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16581
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24425/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18495/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3870/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24988/)

InvincibleRMC:
Another old mypy bug.

```python
import typing

class A:

    def __init__(self) -> typing.NoReturn:   # The return type of "__init__" must be None
        raise NotImplementedError()
```

Will add a # type: ignore for now. Cannot wait for `rhel 10`.

rosidl_generator_py/resource/_action.py.em
```diff
@@ -92,5 +102,5 @@ class @(action.namespaced_type.name)(metaclass=Metaclass_@(action.namespaced_typ
         # The generic message for get the status of a goal.
         from action_msgs.msg._goal_status_array import GoalStatusArray as GoalStatusMessage
 
-    def __init__(self):
+    def __init__(self) -> typing.NoReturn:  # type: ignore
```
christophebedard, line 106: If we have to ignore typing for this line due to a mypy bug, can you mention that (and the mypy bug number) in a comment (maybe right after "ignore") so that we can remove this later? Same for `rosidl_generator_py/resource/_srv.py.em`

InvincibleRMC: It seems like there comment parsing cannot support that. https://mypy-play.net/?mypy=latest&python=3.12&gist=b4d9d618f1831fdb30dc5dcc768a499f I'll add it above for now

christophebedard:
Pulls: ros2/rosidl_python#206
Gist: https://gist.githubusercontent.com/christophebedard/6ea01a76e36980e450b65455f81bcd72/raw/1cb2fe9a7ed734bbd4a459211613e40209a97991/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16607
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24456/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18521/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3900/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25014/)

InvincibleRMC:
`rhel` strikes again 😢

christophebedard:

christophebedard:
Alright, let's try a full CI run again.

christophebedard:
Pulls: ros2/rosidl_python#206
Gist: https://gist.githubusercontent.com/christophebedard/1f6a02ffe82ec8f80bdc504b48fab103/raw/1cb2fe9a7ed734bbd4a459211613e40209a97991/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16612
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24471/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18527/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3929/)
    * Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3942/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25022/)

InvincibleRMC:
@christophebedard the `rhel` CI crashed can you restart it when you get a chance.

christophebedard:
It usually gets automatically retried if it fails due to an infra issue, but it's not always easy to find the corresponding job.

ci_linux-rhel 3922 was retried with 3925, which also failed due to an infra issue and was retried with 3929: https://ci.ros2.org/job/ci_linux-rhel/3929/. There's a `ros2topic` test failure, which is slightly suspicious. It's hard to tell what went wrong, but it could be due to the potential crosstalk we've been seeing. Let's retry it: Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3942/)

christophebedard:
The RHEL job had some different test failures; I don't think it's related to these changes. The first one seems to be Zenoh router-related and the second one seems to be a time-related flake.

christophebedard:
@sloretz do you want to give this another look? Otherwise, I would go ahead and approve + merge it.

christophebedard (approved):

InvincibleRMC:
Thanks so much for all the reviews and finally getting this merged!

christophebedard:
I created a new release (`0.26.0`): https://github.com/ros/rosdistro/pull/47213.

The timing is great, because a [Rolling sync was just completed](https://discourse.openrobotics.org/t/new-packages-for-ros-2-rolling-ridley-2025-08-03/49406), so we'll have time to let the buildfarm rebuild downstream packages and fix anything that breaks before the next sync.

InvincibleRMC:
Feel free to ping me about any downstream problems you find.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -293,15 +349,25 @@ if isinstance(type_, AbstractNestedType):
...
+@[    continue]@
+@[  end if]@
+@{
+import inspect
+import builtins
+noqa_string = ''
+if member.name in dict(inspect.getmembers(builtins)).keys():
+    noqa_string = ', A002'
+}@
+                 @(member.name): typing.Optional[@(type_annotations_setter[member.name])] = None,  # noqa: E501@(noqa_string)
```
peci1, line 364: This pollutes global namespace

rosidl_generator_py/resource/_msg.py.em
peci1, line 391: This pollutes local namespace

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,50 +393,41 @@ if isinstance(type_, AbstractNestedType):
...
 @[      end if]@
 @[    elif isinstance(type_, BasicType) and type_.typename == 'octet']@
-        self.@(member.name) = kwargs.get('@(member.name)', bytes([0]))
+        self.@(member.name) = @(member.name) if @(member.name) is not None else bytes([0])
 @[    elif isinstance(type_, BasicType) and type_.typename in CHARACTER_TYPES]@
-        self.@(member.name) = kwargs.get('@(member.name)', chr(0))
+        self.@(member.name) = @(member.name) if @(member.name) is not None else chr(0)
 @[    else]@
-        self.@(member.name) = kwargs.get('@(member.name)', @(get_python_type(type_))())
+        self.@(member.name) = @(member.name) if @(member.name) is not None else @(get_python_type(type_))()
```
peci1, line 421: This requires polluted local namespace

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -328,50 +393,41 @@ if isinstance(type_, AbstractNestedType):
...
         else:
-            self.@(member.name) = numpy.array(kwargs.get('@(member.name)'), dtype=@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['dtype']))
+            self.@(member.name) = numpy.array(@(member.name), dtype=@(SPECIAL_NESTED_BASIC_TYPES[member.type.value_type.typename]['dtype']))
             assert self.@(member.name).shape == (@(member.type.size), )
 @[        else]@
-        self.@(member.name) = kwargs.get(
-            '@(member.name)',
-            [@(get_python_type(type_))() for x in range(@(member.type.size))]
-        )
+        self.@(member.name) = @(member.name) if @(member.name) is not None else [@(get_python_type(type_))() for x in range(@(member.type.size))]
```
peci1, line 407: This requires polluted local namespace

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -424,19 +480,21 @@ type_ = member.type
...
 if member.name in dict(inspect.getmembers(builtins)).keys():
     noqa_string = '  # noqa: A003'
+
+array_type_commment = ''
+if isinstance(member.type, (Array, AbstractSequence)):
+    array_type_commment = '   # typing.Annotated can be remove after mypy 1.16+ see mypy#3004'
 }@
     @@builtins.property@(noqa_string)
-    def @(member.name)(self):@(noqa_string)
+    def @(member.name)(self) -> @(type_annotations_getter[member.name]):@(noqa_string)@(array_type_commment)
```
peci1, line 492: This requires polluted global namespace

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -424,19 +480,21 @@ type_ = member.type
...
         """Message field '@(member.name)'."""
         return self._@(member.name)
 
     @@@(member.name).setter@(noqa_string)
-    def @(member.name)(self, value):@(noqa_string)
+    def @(member.name)(self, value: @(type_annotations_setter[member.name])) -> None:@(noqa_string)
```
peci1, line 497: This requires polluted global namespace

rosidl_generator_py/resource/_msg.py.em
peci1, line 528: This pollutes local namespace

peci1, line 562: This requires polluted local namespace

peci1, line 622: This requires polluted local namespace

peci1 (commented):
This PR (and not only this one, but this one is the major contributor) imposes new constraints on the names of messages.

I've reported the problem in https://github.com/ros2/rosidl_python/issues/257 . I used the review feature on this PR to highlight the places where I see problems (only in msg.em for simplicity).

PR #211, opened 2024-04-09
Revert install of .so files into python path

RethinkMatze:
There seems that some regression might have happened after #195. When removing those 2 lines, we avoid to install the .so files in lib *and* python path.

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -236,10 +236,6 @@ endforeach()
 
 set_target_properties(${_target_name_lib} PROPERTIES COMPILE_OPTIONS "${_extension_compile_flags}")
 
-set_target_properties(${_target_name_lib} PROPERTIES
-  LIBRARY_OUTPUT_DIRECTORY ${_output_path}
-  RUNTIME_OUTPUT_DIRECTORY ${_output_path})
-
 if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
   install(TARGETS ${_target_name_lib}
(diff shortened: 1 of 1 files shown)


clalancette:
> There seems that some regression might have happened after #195. When removing those 2 lines, we avoid to install the .so files in lib _and_ python path.

Can you explain the regression?  We haven't been seeing that in our testing, so I'm curious what is happening.

RethinkMatze:
Hi @clalancette ,

sure. When you build for example builtin-interfaces, you will get 
`/usr/lib/libbuiltin_interfaces__rosidl_generator_c.so`
*and*
`/usr/lib/python3.12/site-packages/builtin_interfaces/libbuiltin_interfaces__rosidl_generator_py.so`
which are essentially providing the same stuff. The later one is not needed and just bloats the packages.

RethinkMatze:
Note that it was removed in #195 but sneaked in again later with 1505ac07675800c8259ae40df3421177ce0eb17a

clalancette:
> Note that it was removed in #195 but sneaked in again later with [1505ac0](https://github.com/ros2/rosidl_python/commit/1505ac07675800c8259ae40df3421177ce0eb17a)

Ah, thanks for the explanation.  You are right, it did sneak back in.  Here is CI to see if this works everywhere:


* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20829/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15133/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21515/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21516/)

clalancette:
Both of the failing tests here are known issues, so I'm going to go ahead and merge this one.  Thanks for fixing this again.

PR #204, opened 2024-02-15
Remove python_cmake_module and set hints

clalancette:
This must be merged before https://github.com/ros2/ros2/pull/1524 ; see that pull request for more information about this change.

PR #214, opened 2024-07-13
[backport] Use modern cmake targets to avoid absolute paths to appear in binary …

xueying4402:
…archives (#160)

* Use FindPython3

* Make link libraries private when possible to avoid having to export dependencies

* Use Python3::Numpy

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -21,6 +21,7 @@ find_package(PythonInterp 3.6 REQUIRED)
 
 find_package(python_cmake_module REQUIRED)
 find_package(PythonExtra MODULE REQUIRED)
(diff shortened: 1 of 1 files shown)


clalancette (commented):
I have to say that I'm very worried about doing this.  I've been poking at this for months in #204 (though that has a different goal), and the whole system here is tricky.  For a stable release, I think this is going to be dangerous and may have unwanted side-effects.  For those reasons, my opinion is that we should *not* take this into `humble` at this time.

@sloretz What's your take on this one?

sloretz:
Thank you for the PR, and for your patience! I'm going to close this backport for now because I don't feel confident that it [won't break someone's workflow](https://xkcd.com/1172/).

> @sloretz What's your take on this one?

I think my main concern is that Python3 CMake module could find a different version of NumPy than the current custom logic in rosidl_generator_py. I think that could break a non-standard Python setup like a venv. While ROS 2 doesn't directly support those, I've seen bugs opened by people using different non-standard Python setups for other platforms.

PR #216, opened 2024-09-10
Added forced casting of int values ​​to float for float-typed fields

### LINKED ISSUE #172: Integers should be accepted as floats
## Bug report

**Required Info:**

- Operating System: Ubuntu 22.04
- Installation type: binary
- Version or commit hash: 0.15.0
- DDS implementation: cyclonedds
- Client library (if applicable): rclpy

#### Steps to reproduce issue
1. Define a message with a float field
2. Try to assign an integer to the field.

#### Expected behavior
The integer should be automatically casted to a float.

#### Actual behavior
You manually have to cast the integer to float, otherwise you get the error message `AssertionError: The 'x' field must be of type 'float`.

EsipovPA:
Added casting int to floats for float-typed fields in messages
Added test for new feature

Closes [this](https://github.com/ros2/rosidl_python/issues/172) issue

### DIFF
--- a/rosidl_generator_py/CMakeLists.txt
+++ b/rosidl_generator_py/CMakeLists.txt
@@ -53,6 +53,7 @@ if(BUILD_TESTING)
       msg/BuiltinTypeSequencesIdl.idl
       msg/StringArrays.msg
       msg/Property.msg
+      msg/Float.msg
       ADD_LINTER_TESTS
       SKIP_INSTALL
     )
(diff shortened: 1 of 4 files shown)


clalancette (commented):
I have to say that personally, I'm not a fan.  `int` and `float` are not the same thing, and thus I think any conversion like this should be up to the user to explicitly do.  I'd like to hear opinions from @ros2/team, though.

EsipovPA:
Hello @clalancette!

Thanks for the quick reply.

It seems to me that this is a case of a widening conversion, that should be an option for these types of fields. In case of C++ code, the int values would be casted to float automatically, when passed to `=` operator. The idea behind this code is to make possible something like this for python code. Also it just makes python code of message initialization simpler and easier to read, as I see it.

For example, instead of something like this

```
msg = Msg(check_fields=True, value=float(1))
```

or
 
```
msg = Msg(check_fields=True, value=1.)
```

user should be able to write just

```
msg = Msg(check_fields=True, value=1)
```

and get the float value without raised exceptions.

Still confused about the failed test though. Checking

clalancette:
> It seems to me that this is a case of a widening conversion, that should be an option for these types of fields. In case of C++ code, the int values would be casted to float automatically, when passed to `=` operator. The idea behind this code is to make possible something like this for python code. Also it just makes python code of message initialization simpler and easier to read and write, as I see it.

Yeah, I understand what you are trying to do.  I still don't think it is a good idea, even if C++ allows it (there is nothing we can do about that, it is part of the language).

lights0123:
At the minimum, the asserts should be converted to Python exceptions. I set one field of a `Vector3` to 0, and got this error:

```
python3: ./.obj-x86_64-linux-gnu/rosidl_generator_py/geometry_msgs/msg/_vector3_s.c:76: geometry_msgs__msg__vector3__convert_from_py: Assertion `PyFloat_Check(field)' failed.
```

with *no* stacktrace or what the actual type provided was. This took around 20 minutes to debug with gdb and some manual type checking.

I do strongly believe that ints should be implicitly converted—I would argue that using `0` to clear a field is reasonable.

Issue #217, opened 2024-09-27
[CSA Bug Report]Clang Static Analyzer(CSA) Memory Bug Detected


wr-web:
## Bug report

Clang Static Analyzer(CSA), pyrefcon(http://lcs.ios.ac.cn/~maxt/PyRefcon/ASE-2023.pdf)

- Operating System:
  - Linux d18de72e1bb7 5.4.0-196-generic x86

## Bug Type: Access released PyObject/Use After Free
_msg_support.c.em
Commit: https://github.com/ros2/rosidl_python/blob/1fbd99b5fe1fa04674c73e7e5992ccc88e28157c/rosidl_generator_py/resource/_msg_support.c.em#L179-L193
Detail: after Py_DECREF module_attr and class_attr may be released, module_name and class_name are possible freed, Causing Access released PyObject/Use After Free.

Prove of Content(POC):
```c
static *
poc(PyObject * object) {
  PyObject * module_attr = PyObject_GetAttrString(object, "__class__");
  char * module_name = NULL;
  if (module_attr) {
    PyObject *name_attr = PyObject_GetAttrString(module_attr, "__name__");
    if (name_attr) {
      module_name = (char *)PyUnicode_1BYTE_DATA(name_attr);
      Py_DECREF(name_attr);
      Py_DECREF(module_attr);
      printf("%s", module_name);
    }
  }
  return PyLong_FromLong(0);
}
```
And this is the correct result.
<img width="545" alt="image" src="https://github.com/user-attachments/assets/6940f13e-8d9e-47c6-9711-4693c637e96b">

However, If the garbege collect was triggered between `Py_DECREF(object)` and usage of string `module_name`, things will become troublesome.

```c
static *
poc(PyObject * object) {
  PyObject * module_attr = PyObject_GetAttrString(object, "__class__");
  char * module_name = NULL;
  if (module_attr) {
    PyObject *name_attr = PyObject_GetAttrString(module_attr, "__name__");
    if (name_attr) {
      module_name = (char *)PyUnicode_1BYTE_DATA(name_attr);
      Py_DECREF(name_attr);
      Py_DECREF(module_attr);
      call_gc_collect();
      printf("%s\n", module_name);
    }
  }
  return PyLong_FromLong(0);
}

void call_gc_collect() {
    PyObject *gc_module = PyImport_ImportModule("gc");
    if (gc_module) {
        PyObject *gc_collect = PyObject_GetAttrString(gc_module, "collect");
[...]

PR #218, opened 2024-10-02
Revamp how we check for the correct class.

clalancette:
In an issue it was pointed out that we are decref'ing before we actually use some of the data, which means that in theory the garbage collector could reclaim the data before we used it.  So the original point of this change was to fix that issue.

However, while looking at it I realized we could slightly improve performance here by avoiding a copy of the class and module into a combined string.  Instead, we can compare them separately, which should reduce the copies.

This is a partial solution to https://github.com/ros2/rosidl_python/issues/217 ; the other reported bug in there will require a much more extensive rework.

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.c.em
+++ b/rosidl_generator_py/resource/_msg_support.c.em
@@ -158,41 +158,40 @@ PyObject * @('__'.join(type_.namespaces + [convert_camel_case_to_lower_case_unde
 
 @{
 module_name = '_' + convert_camel_case_to_lower_case_underscore(interface_path.stem)
+class_module = '%s.%s' % ('.'.join(message.structure.namespaced_type.namespaces), module_name)
+namespaced_type = message.structure.namespaced_type.name
 }@
 ROSIDL_GENERATOR_C_EXPORT
 bool @('__'.join(message.structure.namespaced_type.namespaces + [convert_camel_case_to_lower_case_underscore(message.structure.namespaced_type.name)]))__convert_from_py(PyObject * _pymsg, void * _ros_message)
 {
-@{
-full_classname = '%s.%s.%s' % ('.'.join(message.structure.namespaced_type.namespaces), module_name, message.structure.namespaced_type.name)
-}@
   // check that the passed message is of the expected Python class
   {
-    char full_classname_dest[@(len(full_classname) + 1)];
-    {
-      char * class_name = NULL;
-      char * module_name = NULL;
-      {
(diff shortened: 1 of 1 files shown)


clalancette:
Pulls: https://github.com/ros2/rosidl_python/pull/218
Gist: https://gist.githubusercontent.com/clalancette/1319b32b184fce38189666d4e4e69c43/raw/2b4fb3b5d7b2c647a8f6fe22b9010efcd374125e/ros2.repos
BUILD args:  --packages-above-and-dependencies ament_cmake_core
TEST args:  --packages-above ament_cmake_core
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14647
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21930/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16197/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1444/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22656/)

clalancette:
I had to add a fix when building in Release mode.  Another set of CI is upcoming.

clalancette:
Pulls: https://github.com/ros2/rosidl_python/pull/218
Gist: https://gist.githubusercontent.com/clalancette/1319b32b184fce38189666d4e4e69c43/raw/2b4fb3b5d7b2c647a8f6fe22b9010efcd374125e/ros2.repos
BUILD args:
TEST args:
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14647
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21936/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16203/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1450/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22662/)

sloretz (approved):
I don't see any issues, so LGTM!

If I were going to rewrite this module from scratch I'd make this check use [PyObject_IsInstance](https://docs.python.org/3/c-api/object.html#c.PyObject_IsInstance) instead of using string comparisons, but that'd require importing the class type we want to check, which we'd probably want to do once in a `PyInit_*()`, and I'm not sure where the `PyInit_*()` function even is.

PR #237, opened 2025-10-01
Extract imports out from `__init__()` and `__type_support__()`

InvincibleRMC:
## Description

For about a `5%` speed up in message construction times and simplification of importing logic by moving imports out of `__init__()` and  `__type_support__()` .

All part of a grand plan to maybe support `Optional` Idl types by further simplify the `__init__()` method.

No only internals should change.


peci1:
Related to https://github.com/ros2/rosidl_python/issues/257 , importing the message type names into global namespace is dangerous and should be avoided. It would be better to import just the package and use fully qualified names.

PR #246, opened 2026-02-01
Refactor Check Fields into helper function

InvincibleRMC:
## Description

For the work around making sequences cast to list it required several changes to the spaghetti which was the template around it. This PR only moves the spaghetti out into python where making these changes is easier to test going forward and easier to enforce coding standards.  A big of the reason this was done is the upcoming draft of the PR I had it basically required running this block a second time in some cases and the way it was templated it just was not possible without just copying the whole block and hoping.

I tested this make no difference by running `diff -r` on the messages generated by base rolling and this branch to ensure there were no difference.

No It should only be an internal design detail.


InvincibleRMC:
@christophebedard would you be able to review this?

christophebedard:
I'm not a big fan of doing (basically) template expansion in Python files directly; I think in general we should be moving the other way, to `.em` files. Is there another way to keep the code in _an_ `.em` file and still be able to re-use it? Maybe have a separate `.em` file that we can include multiple times? I'm not very familiar with this

christophebedard:
I believe this was superseded by #249, so I'll close this.

PR #251, opened 2026-03-19
Fix issue #1679: Add deprecation of interface variables

Pengkun-ZHU:
Companion to ros2/rosidl#946 (fixes ros2/ros2#1679)

## Summary

Adds `DeprecationWarning` support to the Python message generator for fields annotated with `@deprecated`.

### `rosidl_generator_py/resource/_msg.py.em`

For any field whose IDL `Member` carries a `@deprecated` annotation (set via `# @deprecated` in the `.msg` source), the generated Python property getter and setter now emit:

```python
import warnings
warnings.warn(
    "field 'old_field' in 'MyMessage' is deprecated",
    DeprecationWarning,
    stacklevel=2)
```

This fires at runtime whenever user code reads or writes the deprecated field, following standard Python deprecation convention (`stacklevel=2` points the warning at the caller, not the generated code).

## Coverage

The `_msg.py.em` template is the single leaf template invoked for all message-like types — messages, service request/response/event sub-messages, and action goal/result/feedback sub-messages — so the warning propagates automatically to `.srv` and `.action` fields as well.

## Known limitations / out of scope
- **Whole-message deprecation** is not addressed here; only field-level.
- **No custom deprecation message** — `# @deprecated` carries no free-text argument in this version.
- **Wire format is unchanged** — `@deprecated` does not affect type hash or serialization.

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -431,11 +431,25 @@ if member.name in dict(inspect.getmembers(builtins)).keys():
 }@
     @@builtins.property@(noqa_string)
     def @(member.name)(self):@(noqa_string)
+@[  if member.has_annotation('deprecated')]@
+        import warnings
+        warnings.warn(
+            "field '@(member.name)' in '@(message.structure.namespaced_type.name)' is deprecated",
+            DeprecationWarning,
+            stacklevel=2)
+@[  end if]@
         """Message field '@(member.name)'."""
         return self._@(member.name)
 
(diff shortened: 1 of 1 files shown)


InvincibleRMC:
I can write my own implementation or you can update yours. Either is fine but you would have to add the deprecation push pop probably to the message.c.em. You should also use deprecation from typing_extensions. And finally use the text from the annotation some where in the message so user can know what to migrate to

Pengkun-ZHU:
Sure, I will make a further change to align with your implementation.

PR #256, opened 2026-04-03
Add Python generator support for deprecated interface fields

Pengkun-ZHU:
Adds support for the `@deprecated` IDL annotation in the Python generator rosidl_generator_py.

When a message field is annotated with `@deprecated`, the generated Python class will:
- Emit a DeprecationWarning (via PEP 702 `typing_extensions.deprecated`) when the field is accessed or set via the public property
- Suppress the warning in internal operations (`__eq__`, `__repr__`, C extension type conversion) to avoid false positives

Depends on: ros2/rosidl#945

Signed-off-by: Pengkun-ZHU <q1091803103@gmail.com>

### DIFF
--- a/rosidl_generator_py/CMakeLists.txt
+++ b/rosidl_generator_py/CMakeLists.txt
@@ -49,6 +49,7 @@ if(BUILD_TESTING)
       msg/BuiltinTypeSequencesIdl.idl
       msg/StringArrays.msg
       msg/Property.msg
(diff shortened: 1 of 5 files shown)


asymingt:
We're quite close to feature freeze for ROS 2 Lyrical. Any chance we can re-target this to M-Turtle?

InvincibleRMC:
Yeah this is fine for m turtle.

PR #250, opened 2026-03-16
Add support for rosidl::Buffer in rosidl Python path for rclpy

nvcyc:
## Description

This pull request adds `rosidl::Buffer` support for uint8[] fields in the Python code generation path.

This pull request consists of the following key changes:

- **Python message setters** (`_msg.py.em`): For `uint8[]` fields, the generated setters now allow assigning from `rosidl_buffer.Buffer` or plain `array.array`. Existing code that sets `msg.data = array.array('B', ...)` continues to work unchanged.
- **Python/C conversion** (`_msg_support.c.em`): Messages with `uint8[]` fields use the sequence's `is_rosidl_buffer` flag to determine the conversion path:
  - **Python to C**: For non-CPU `rosidl_buffer.Buffer` objects, calls `_get_buffer_ptr()` and sets the C++ Buffer pointer plus `is_rosidl_buffer = true` so the RMW path uses the existing buffer without copying. For `array.array` objects, uses the normal byte-copy path.
  - **C to Python**: When the RMW deserializes into a vendor-backed buffer (`is_rosidl_buffer == true`), calls `_take_buffer_from_ptr()` to wrap the pointer in a Python `Buffer` (taking ownership) and set it on the message. CPU-based data is converted to `array.array` as before.

### Is this user-facing behavior change?

This pull request does not change existing rclpy behavior. Existing code using `array.array` for `uint8[]` fields continues to work. The `rosidl_buffer.Buffer` type is only encountered when a subscriber receives data from a non-CPU buffer backend (and has opted in via `acceptable_buffer_backends`).

### Did you use Generative AI?

Yes. Claude (claude-4.6-opus) via Cursor was used to assist with creating an initial prototype version of the changes contained in this PR.

### Additional Information

This PR is part of the broader ROS 2 native buffer feature introduced in this [post](https://discourse.openrobotics.org/t/working-prototype-of-native-buffers-accelerated-memory-transport/52399).

PR #263, opened 2026-08-12
Add testcase and fix for include order of generated code with stdint.h

mathias-luedtke:
## Description

This PR add a test case that shows that cpplint checks fail for generated code for services that have buffers in their response, but not in the request.
The same issue should happen for actions.

The tests will fail with:
```
- rosidl_generator_py.cpplint_rosidl_generated_py build/include_order [4] (/ws/build/rosidl_generator_py/rosidl_generator_py/rosidl_generator_py/srv/_get_buffer_s.c:83)
  <<< failure message
    Found C system header after other header. Should be: _get_buffer_s.h, c system, c++ system, other.
  >>>
```

No

### Did you use Generative AI?

No

### Additional Information

This issue was introduced by https://github.com/ros2/rosidl_python/pull/250


rosidl_generator_py/srv/GetBuffer.srv
```diff
@@ -0,0 +1,2 @@
+---
```
fujitatomoya, line 1: the file name should be `GetBuffer.srv`? please make sure build and test on your local environment when you create the PR, this is really easy to notice.

skyegalaxy:
@adbidwai to take a look

adbidwai:
Hi, I can take a look at this. Thanks!

fujitatomoya (changes requested):
I do not think we can take this without fixes, because this generates intentionally red test.

rosidl_generator_py/srv/GetBuffer.srv
mathias-luedtke: My bad!
I fixed it locally, but forgot to commit the rename.

mathias-luedtke:
> I do not think we can take this without fixes, because this generates intentionally red test.

I can add the fix as well, but I need to know which option you do prefer (see https://github.com/ros2/rosidl_python/issues/264)

mathias-luedtke:
@fujitatomoya: Which fix would you prefer?

* silence build/include_order in the cpplint checks
* always include <stdint.h>
* more complicated fix: add stdint.h in the request/goal part, if the parts use buffers

fujitatomoya:
I'd say that the best option we got here is to hoist the include to the top-of-file system-header block, conditionally at the interface level with checking that interface requires the header. but i guess that perfectly acceptable fallback: always include <stdint.h> in the top block if that makes the complicatoin. I was thinking about the compile-time concern but stdint.h is one of the smallest headers in the standard library. what do you think?

mathias-luedtke:
> I'd say that the best option we got here is to hoist the include to the top-of-file system-header block, conditionally at the interface level with checking that interface requires the header.

I came up with https://github.com/mathias-luedtke/rosidl_python/commit/620a02674057d068abbf20c22b1baf92f343faeb
However, I am not convinced..

> but i guess that perfectly acceptable fallback: always include <stdint.h>

It turns out that Python.h already includes it, so it should not affect compile time.
-> We could even drop the stdint.h include and it still passes.

Furthermore - depending on the message type - the generated code already uses the types from stdint.h without including them explicitly.

IMHO always including `stdint.h` is the better fix: it is simple and maintainable.
And it allows us to drop the has_buffer_fields handling, which should make up for any (negligible) include overhead.

MatthijsBurgh:
For whatever it is worth for the re-review: I checked 9dbae598 against a real downstream failure and it resolves it.

`ed_gui_server_interfaces/srv/GetEntityInfo.srv` has exactly the shape from #264 — no buffer in the request, two in the response — and is red on rolling:

```
- ed_gui_server_interfaces.cpplint_rosidl_generated_py build/include_order [4] (.../srv/_get_entity_info_s.c:137)
  <<< failure message
    Found C system header after other header. Should be: _get_entity_info_s.h, c system, c++ system, other.
  >>>
```

(https://github.com/tue-robotics/ed_gui_server/actions/runs/33599833100)

I have no rolling environment at hand, so I checked it two ways:

- The include-emission region of `_msg_support.c.em` on this branch — from `include_parts = ` down to `// end nested array functions include` — is identical to jazzy's apart from the `'stdint.h'` entry and an empy comment. Jazzy generates a cpplint-clean `_get_entity_info_s.c` for this service, so this branch does too.
- `ament_cpplint --filters=-readability/fn_size --linelength 999` over the generated file for that service: clean as jazzy generates it, reproduces the error above with `<stdint.h>` live in the response block as rolling emits it today, clean again with `<stdint.h>` in the top block as this branch emits it.

The second is on a reconstruction of the rolling output rather than the real thing, so please read it as corroboration of the first rather than an independent result. The middle case is the control — without it a clean run would not mean much.

The same argument covers the `GetBuffer.srv` added here: with `stdint.h` unconditional it is always in the first block, so the testcase should be red before 9dbae598 and green after, which I think addresses the "generates intentionally red test" point from the review.

